PR #11, opened 2019-09-17
Added support for XPath selectors

pjforde1978:
I am excited to propose the addition of support for XPath selectors in addition to CSS selectors to the cable_ready API. This will facilitate some exciting features on related projects such as stimulus_reflex.

This change is non-breaking, as it still expects a CSS selector by default.

However, if you happen to pass in an XPath query that resolves to a single DOM element via config.selector, you can set the new config.xpath boolean option to true and it will handle it no problem.


hopsoft:
This is great. I'm happy to accept but ask that you first merge master and then run `bin/standardize`. Also that you add a section about using XPath to the readme. Thanks.

hopsoft:
🤦‍♂ Looks like I created a conflict for you. Sorry about that.

bin/standardize
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 bundle exec standardrb --fix
-cd ./javascript && yarn run prettier --write cable_ready.js
+cd ./javascript && yarn run prettier --write cable_ready.js
```
pjforde1978: I don't understand how this is different. I don't think the file had a newline before?

pjforde1978:
Even though Code Climate disagrees, I believe that my changes to the JS library and the README are ready for review. StandardRB did raise some warnings but they were idiomatic and would require breaking your API.

PR #18, opened 2019-10-27
cancel morph if numeric input is invalid

leastbad:
covers https://bugzilla.mozilla.org/show_bug.cgi?id=1398528


hopsoft:
I like this idea but think this PR may be solving the problem in the wrong location. It's assuming invalid numeric data coming from the server (which should be a more trustworthy source of data). Seems more appropriate to prevent triggering a reflex on the StimulusReflex side of things so invalid values aren't passed to the server in the first place. Thoughts?

pjforde1978:
I am fine with this. The reason this PR exists is that in the fall i did a
deep-dive audit of LiveView looking for potential future pain points that
we could prevent based on their presumably hard-won lived experience - I
put a lot of stock in Jose’s time on Rails core.

Best as I can tell, this was the other side of the invalid number
protection that is part of the original SSoT PR.

So yeah: I’m honestly not sure how neurotic this is.

We *could* just ask them...

PR #23, opened 2020-01-21
Add support to ignore missing DOM elements via options parameter

pjforde1978:
While it is desirable to emit console log messages when an invalid DOM selector is received in development, in production there should be an escape hatch to suppress messages if a valid reason exists to do so.

This PR introduces an optional 2nd parameter to the perform method, an object called options. Currently, this object has a single key  `processMissing` which defaults to true. If set to false, DOMOperations will not be called unless there is a valid element in scope.

I added a documentation hint and supplemented the original catch block with a distinction to alert developers to the missing DOM element. I spent a lot of time trying to figure out what was going wrong before realizing that the element was missing due to an unhelpful default error message.

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -158,7 +158,7 @@ const DOMOperations = {
   }
 }
 
-const perform = operations => {
+const perform = (operations, options = { processMissing: true }) => {
   for (let name in operations) {
     if (operations.hasOwnProperty(name)) {
       const entries = operations[name]
@@ -172,9 +172,13 @@ const perform = operations => {
           } else {
             detail.element = document
           }
-          DOMOperations[name](detail)
+          if (detail.element || options.processMissing)
+            DOMOperations[name](detail)
         } catch (e) {
-          console.log(`CableReady detected an error in ${name}! ${e.message}`)
+          if (entries[i].element)
+            console.log(`CableReady detected an error in ${name}! ${e.message}`)
+          else
+            console.log(`CableReady ${name} failed due to missing DOM element.`)
         }
       }
     }
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -158,7 +158,10 @@ const DOMOperations = {
   }
 }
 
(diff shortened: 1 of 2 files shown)


hopsoft:
I'm still thinking about this one. I'm not opposed to the feature but may have you change the implementation a little. Will get back to you on this soon. Thanks.

pjforde1978:
I’m not attached to how it gets done so long as my devs won’t see confusing
error logs via normal usage.

I’m using CR differently than we have in the past, and to great effect. Up
until now we’ve been in a position to design all apps to only emit messages
we know will work.

The spirit of this PR is to be liberal in what we accept.

javascript/cable_ready.js
```diff
@@ -158,7 +158,7 @@ const DOMOperations = {
   }
 }
 
-const perform = operations => {
+const perform = (operations, options = { processMissing: true }) => {
```
hopsoft: I like the idea of the options hash, but would prefer something like this for the name and default value. 

```js
{ignoreMissingElements: false}
```

javascript/cable_ready.js
```diff
@@ -172,9 +172,13 @@ const perform = operations => {
           } else {
             detail.element = document
           }
-          DOMOperations[name](detail)
+          if (detail.element || options.processMissing)
+            DOMOperations[name](detail)
```
hopsoft:
```js
if (!options.ignoreMissingElements) DOMOperations[name](detail)
```

The negated check doesn't read quite as well here, but I think the new option name and default value make more sense to consumers of the lib. What do you think?

pjforde1978:
I have never encountered a double negative that was more readable, come on.
:)

I think the sanest thing to do is use a more descriptive key that,
remember, newcomers might not need until they are far enough along to get
it.

I counter-propose `raiseMissingElementWarnings = true`. This is much more
descriptive at the expense of a few more characters - but it keeps the gate
logic much simpler.

hopsoft:
I like it. Let's run with `emitMissingElementWarnings`.

pjforde1978:
Awesome! I'll update and push.

PR #26, opened 2020-02-26
dispatchEvent was passing non-existant name var

leastbad:


leastbad:
Please disregard. Embarrassing mistake on my end. Sorry for the distraction.

PR #28, opened 2020-03-05
Set cookie support

hopsoft:


jonathan-s:
If I were a user of the set_cookie method in ruby. I'd prefer the method to have the signature `
```
set_cookie(key, value, expires=None, max_age=None)
```

That way you don't need to remember how the cookie should be formatted.

PR #34, opened 2020-04-29
add set_property method to library

leastbad:
A marriage of set_attribute and set_value...

While it is true that 90% of the time you want to use set_attribute or set_value, there are scenarios in which a developer might want to set a property on an element.

This implementation runs a simple check to verify that a given property name is present in the element's property dictionary before setting the value. It does **not** attempt to verify that the developer has passed a value that is compatible with the property's data type. If someone tries to set an invalid data type, it will be caught by CR's standard try/catch error handling.


leastbad:
My only hangup is that I don’t love contributing code i hope not to need, but from a completionist perspective, this is technically missing.

PR #35, opened 2020-05-05
Fix typo in method name outerHTML -> outer_html

back2war:


leastbad:
Nice catch, @back2war!

PR #37, opened 2020-05-14
Add setStyles

excid3:
This adds a `setStyles` operation that allows for setting multiple style attributes at once on an element. This actually loops through and delegates each entry to the `setStyle` operation.

For example, we can set background and color at once:

```ruby
cable_ready["MyChannel"].set_styles(
  selector: "string", # required - string containing a CSS selector or XPath expression
  styles: {
    background: "red",
    color: "white"
  }
)
```

Depending on browser compatibility targets / goals, we could shorten this by one line using `Object.entries` instead. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -235,6 +235,15 @@ const DOMOperations = {
     dispatch(element, 'cable-ready:after-set-style', detail)
   },
 
+  setStyles: detail => {
+    const { element, styles } = detail
+    for (const property in styles) {
+      const name = property,
+        value = styles[property]
+      DOMOperations['setStyle']({ element: element, name: name, value: value })
+    }
+  },
+
   // Dataset Mutations .......................................................................................
 
   setDatasetProperty: detail => {
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -237,11 +237,11 @@ const DOMOperations = {
 
   setStyles: detail => {
     const { element, styles } = detail
-    for (const property in styles) {
-      const name = property,
-        value = styles[property]
-      DOMOperations['setStyle']({ element: element, name: name, value: value })
+    dispatch(element, 'cable-ready:before-set-styles', detail)
+    for (let [name, value] of Object.entries(styles)) {
+      element.style[name] = value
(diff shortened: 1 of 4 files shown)


javascript/cable_ready.js
```diff
@@ -235,6 +235,15 @@ const DOMOperations = {
...
 
+  setStyles: detail => {
+    const { element, styles } = detail
+    for (const property in styles) {
+      const name = property,
+        value = styles[property]
+      DOMOperations['setStyle']({ element: element, name: name, value: value })
+    }
+  },
+
```
hopsoft, line 246: Should we dispatch before/after events for this?

hopsoft:
I also like the idea of updating to use `Object.entries`

javascript/cable_ready.js
leastbad: Won't those events be generated by the setStyle operations this creates?

leastbad: Unless you see value in highlighting that several values were updated in addition to the events generated by the individual updates. I feel like it's a bit much, but I'm listening!

hopsoft: It's definitely a lot of events, but it would be consistent. Also, wondering if we should suppress the individual events when `setSetyles` is used and only emit for the batch operation. I'm curious to see what @excid3 thinks.

excid3: Great questions. I feel like it'd be good to be consistent and have events for `setStyles`. 

We could also skip the delegation to `setStyle` if firing those individual events is a problem. 

I don't personally have a use for those events, so I'm not sure I've got the best opinion on those.

excid3:
I'll update that to use Object.entries as well. 👍

excid3:
I've updated this:

1. Using `Object.entries` now
2. Added events
3. Removed delegation to `setStyle` and just update styles in this operation itself

leastbad:
@excid3 I appreciated the cleverness of the initial solution, but agree that keeping them separated is likely the smart move. Looks good to me!

PR #38, opened 2020-05-17
broadcast_to method

leastbad:
CableReady::Channels exposes a public `broadcast_to(resource)` method, which is used to broadcast to an ActionCable channel which makes use of the `stream_for` method. When using `broadcast_to`, we pass the channel constant instead of a String-based identifier.

```ruby
class DocumentChannel < ApplicationCable::Channel
  def subscribed
    stream_for Document.find(params[:document_id])
  end
end

class Document< ApplicationRecord
  after_save do
    cable_ready[DocumentChannel].morph({selector: "#document", html: self.content})
    cable_ready.broadcast_to(self)
  end
end
```

A resource must be passed to the broadcast_to method. ActionCable builds a subscription handle based on the name of the channel and a unique hash of the resource.

When using CableReady, it is possible to have multiple identifiers "on the go" at one time, collecting operations to be run.

Both `broadcast` and `broadcast_to` methods to support an optional identifiers parameter **splat** (h/t: @StephenFiser). Default behaviour is unchanged from current behaviour, which is that all channels are broadcast one at a time. You can pass zero, one or several identifiers.

```ruby
cable_ready.broadcast_to User.find(1), FooChannel, BarChannel
```

`broadcast` will ignore Channel constant identifiers. `broadcast_to` will ignore String identifiers. This means you can have String and Channel identifiers "on the go" at the same time, and they will not conflict.

```ruby
cable_ready['test1'].morph()
cable_ready['test2'].morph()
cable_ready[FooChannel].morph()
cable_ready[BarChannel].morph()

cable_ready.broadcast # broadcasts to test1 and test2
cable_ready.broadcast('test1') # broadcasts to test1
cable_ready.broadcast(FooChannel) # ignored

cable_ready.broadcast_to(Foo.find(1)) # broadcasts to foo:Z2lkOi8vdGhlLWRlcGFubmV1ci9Qcm9wb3NhbC8x AND bar:Z2lkOi8vdGhlLWRlcGFubmV1ci9Qcm9wb3NhbC8x which is valid but probably not what you want
cable_ready.broadcast_to(Foo.find(1), FooChannel) # broadcasts to foo:Z2lkOi8vdGhlLWRlcGFubmV1ci9Qcm9wb3NhbC8x
cable_ready.broadcast_to(Foo.find(1)) # broadcasts to foo:Z2lkOi8vdGhlLWRlcGFubmV1ci9Qcm9wb3NhbC8x AND bar:Z2lkOi8vdGhlLWRlcGFubmV1ci9Qcm9wb3NhbC8x which is valid but probably not what you want
cable_ready.broadcast_to(Foo.find(1), FooChannel, BarChannel) # broadcasts to foo:Z2lkOi8vdGhlLWRlcGFubmV1ci9Qcm9wb3NhbC8x AND bar:Z2lkOi8vdGhlLWRlcGFubmV1ci9Qcm9wb3NhbC8x which is valid but probably not what you want
cable_ready.broadcast_to('test1') # ignored
```

Finally, both `broadcast` and `broadcast_to` now have an optional named last parameter called `clear`, which defaults to `true`. Setting it to `false` prevents CableReady from clearing the channel operations data after a broadcast, allowing it to potentially be sent multiple times or to multiple resources.

```ruby
cable_ready[ImageChannel].morph()
cable_ready.broadcast_to(image1, clear: false)
cable_ready.broadcast_to(image2) # broadcasts the same morph to image2
```

### DIFF
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -2,7 +2,7 @@
 
 module CableReady
   class Channel
-    attr_reader :name, :operations
+    attr_reader :identifier, :operations
 
     # Example Operations Payload:
     #
@@ -119,8 +119,8 @@ module CableReady
     #     value:    "string"
     #   }, ...],
     # }
-    def initialize(name)
-      @name = name
+    def initialize(identifier)
+      @identifier = identifier
       @operations = stub
     end
 
@@ -131,14 +131,14 @@ module CableReady
     def broadcast
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
-      ActionCable.server.broadcast name, "cableReady" => true, "operations" => operations
+      ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
       clear
     end
 
-    def broadcast_to(channel, model)
+    def broadcast_to(model)
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
-      channel.constantize.broadcast_to model, "cableReady" => true, "operations" => operations
+      identifier.broadcast_to model, "cableReady" => true, "operations" => operations
       clear
     end
 
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -124,22 +124,22 @@ module CableReady
       @operations = stub
     end
 
-    def clear
+    def reset
       @operations = stub
     end
 
-    def broadcast
+    def broadcast(clear)
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
       ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
-      clear
+      reset if clear
     end
 
-    def broadcast_to(model)
+    def broadcast_to(model, clear)
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
       identifier.broadcast_to model, "cableReady" => true, "operations" => operations
-      clear
+      reset if clear
     end
 
     def set_cookie(value)
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -127,6 +127,13 @@ module CableReady
       clear
     end
 
+    def broadcast_to(channel, model)
+      operations.select! { |_, list| list.present? }
+      operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
+      channel.constantize.broadcast_to model, "cableReady" => true, "operations" => operations
+      clear
+    end
+
     def set_cookie(value)
       add_operation(:set_cookie, {cookie: value})
     end
--- a/lib/cable_ready/channels.rb
+++ b/lib/cable_ready/channels.rb
@@ -12,23 +12,23 @@ module CableReady
       @channels[identifier] ||= CableReady::Channel.new(identifier)
     end
 
-    def broadcast(*identifiers)
+    def broadcast(*identifiers, clear: true)
       @channels.values
         .reject { |channel| identifiers.any? && identifiers.exclude?(channel.identifier) }
         .select { |channel| channel.identifier.is_a?(String) }
         .tap do |channels|
-          channels.each(&:broadcast)
-          channels.each { |channel| @channels.except!(channel.identifier) }
+          channels.each { |channel| @channels[channel.identifier].broadcast(clear) }
+          channels.each { |channel| @channels.except!(channel.identifier) if clear }
         end
     end
 
-    def broadcast_to(model, *identifiers)
+    def broadcast_to(model, *identifiers, clear: true)
       @channels.values
         .reject { |channel| identifiers.any? && identifiers.exclude?(channel.identifier) }
         .reject { |channel| channel.identifier.is_a?(String) }
         .tap do |channels|
-          channels.each { |channel| @channels[channel.identifier].broadcast_to(model) }
-          channels.each { |channel| @channels.except!(channel.identifier) }
+          channels.each { |channel| @channels[channel.identifier].broadcast_to(model, clear) }
+          channels.each { |channel| @channels.except!(channel.identifier) if clear }
         end
     end
   end
--- a/lib/cable_ready/channels.rb
+++ b/lib/cable_ready/channels.rb
@@ -16,23 +16,19 @@ module CableReady
       @channels = {}
     end
 
-    def broadcast(identifier = nil)
-      if identifier
-        @channels[identifier].broadcast
-      else
-        @channels.values.select { |channel| channel.identifier.is_a?(String) }.each(&:broadcast)
-      end
+    def broadcast(*identifiers)
+      @channels.values
+        .reject { |channel| identifiers.any? && identifiers.exclude?(channel.identifier) }
+        .select { |channel| channel.identifier.is_a?(String) }
+        .each(&:broadcast)
       clear
     end
 
-    def broadcast_to(model, identifier = nil)
-      if identifier
-        @channels[identifier].broadcast_to model
-      else
-        @channels.values.reject { |channel| channel.identifier.is_a?(String) }.each do |channel|
-          @channels[channel.identifier].broadcast_to model
-        end
-      end
+    def broadcast_to(model, *identifiers)
+      @channels.values
+        .reject { |channel| identifiers.any? && identifiers.exclude?(channel.identifier) }
+        .reject { |channel| channel.identifier.is_a?(String) }
+        .each { |channel| @channels[channel.identifier].broadcast_to model }
       clear
     end
   end
--- a/lib/cable_ready/channels.rb
+++ b/lib/cable_ready/channels.rb
@@ -14,8 +14,19 @@ module CableReady
       @channels = {}
     end
 
-    def broadcast
-      @channels.values.map(&:broadcast)
+    def broadcast(channel = nil)
+      channel ? @channels[channel].send(:broadcast) : @channels.values.each(&:broadcast)
+      clear
+    end
+
+    def broadcast_to(model, channel = nil)
+      if channel
+        @channels[channel].send(:broadcast_to, channel, model)
+      else
+        @channels.keys.each do |key|
+          @channels[key].send(:broadcast_to, key, model)
+        end
+      end
       clear
     end
   end
--- a/lib/cable_ready/channels.rb
+++ b/lib/cable_ready/channels.rb
@@ -8,25 +8,29 @@ module CableReady
       @channels = {}
     end
 
-    def [](channel_name)
-      @channels[channel_name] ||= CableReady::Channel.new(channel_name)
+    def [](identifier)
+      @channels[identifier] ||= CableReady::Channel.new(identifier)
     end
 
     def clear
       @channels = {}
     end
 
-    def broadcast(channel = nil)
-      channel ? @channels[channel].send(:broadcast) : @channels.values.each(&:broadcast)
+    def broadcast(identifier = nil)
+      if identifier
+        @channels[identifier].send(:broadcast)
+      else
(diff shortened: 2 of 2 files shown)


hopsoft:
This is pretty great. I'm looking forward to testing it.

hopsoft:
@leastbad Will you add some usage examples to the comments on this PR so we have a paper trail for everyone to follow? It will also help shape the thinking around how to best document this new feature.

leastbad:
I spent a lot of time thinking about what additional information and ultimately, examples would be most helpful. Working with @stephenfiser on #42 helped me clarify what's missing or non-obvious.

In the same way that a lot of people were inspired seeing how you use CableReady broadcasts inside Reflex actions, I believe that there are similarly conceptual blocks keeping people from imagining new patterns and approaches with AC and CR that are a good example away from a lot of people suddenly being really excited about a new way to design these sorts of interfaces.

1. Assuming that all of your channel subscriptions are all importing the same memoized `consumer.js` there appears to be little if any overhead on the client or server to being subscribed to an arbitrary number of streams. Today, most folks using AC are thinking of channels and subscriptions as 1:1. The power of 1:N is still a "locked achievement" in gaming terms. I want to show how clever use of Stimulus' MutationObserver-powered `connect()` works really well with people subscribing to multiple instances of a channel; broadcast_to is key to unlocking that thinking.

2. The `broadcast_to` method that is inherited by every Channel class is really just the entry-point for a useful abstraction. Technically, you could call the same non-public resource serializer method that broadcast_to calls and create a subscription channel with the same identifier token using just `stream_from`. Ultimately it's the same data being sent the same way. However the AC framework has useful tooling (even down to the logging) which differentiates between generalized `broadcast` and resource-specific `broadcast_to`. It makes good sense to continue the same abstraction pattern started by AC by providing a `broadcast_to` in CR that expects a model.

I'm going to offer two examples which are really use cases for this approach.

## Always up-to-date article

If you've read documentation hosted on GitBook, you've probably noticed that the document you're reading will change in real-time as the author makes changes to it. While in this case it's after-commit, there's no particular reason that this approach couldn't stream the current state of the document while the author is typing into it, similar to the real-only mode on Google Docs.

I usually recommend that people don't call CableReady in ActiveRecord callbacks directly, prefering instead of execute ActionJobs. However, I want to keep this light and tight, so please use your imagination.

```ruby
class DocumentChannel < ApplicationCable::Channel
  def subscribed
    stream_for Document.find(params[:document_id])
  end
end
```

```ruby
class Document< ApplicationRecord
  after_save do
    cable_ready['document'].morph({selector: "#content", html: self.content})
    cable_ready.broadcast_to(self)
  end
end
```

```js
import consumer from './consumer'

  connect () {
    consumer.subscriptions.create(
      {
        channel: 'DocumentChannel',
        document_id: this.element.dataset.documentId
      },
      {
        received (data) {
          if (data.cableReady) CableReady.perform(data.operations)
        }
      }
    )
  }
}
```

```html
<div data-controller="document" id="document" data-document-id="<%= @document.id %>">
  <%= @document.content %>
</div>
```

**More to follow**

leastbad:
One of my go-to techniques is to eschew the entire `app/javascript/channels` folder entirely, opting to move `consumer.js` to a neutral location and create Stimulus controllers that are also ActionCable consumers. This allows me to have access to the Stimulus ecosystem, including not just the controller instance but a direct path to `this.application` and of course, the opportunity to initiate StimulusReflex activity.

When combined with the CableReady `broadcast_to` method, we can move past the familiar 1:1 associations we see in a resource view such as Article id#1 and start to imagine UIs composed of many instances of a combined Stimulus+consumer class - including elements that have been injected into the page dynamically.

In other words, once you have a connection established, it's now easily possible to present a vast collection of live components where each is subscribed to its own identifier but all data is still moving via the same channel, which itself could be one of several channels sharing the same consumer (and single WS connection to the server). This is incredibly powerful, as it cuts to the core value proposition that React was created to solve: how do you deliver a newsfeed where every element subscribes to its own feed of likes and comments? Well, with Stimulus and CableReady#broadcast_to, it's a few dozen lines of code.

## Collection of images which display real-time voting

```html
<div data-controller="image-grid">
  <% @images.each do |image| %>
  <div id="image_<%= image.id %>" data-controller="image" data-image-id="<%= image.id">
    <img src="<%= image.path %>">
    <span data-reflex="click->ImageReflex#like"><%= images.likes %></span
  </div>
  <% end %>
</div>
```

```js
import consumer from './consumer'

  connect () {
    consumer.subscriptions.create(
      {
        channel: 'ImageChannel',
        image_id: this.element.dataset.imageId
      },
      {
        received (data) {
          if (data.cableReady) CableReady.perform(data.operations)
        }
      }
    )
  }
}
```

```ruby
class ImageChannel < ApplicationCable::Channel
  def subscribed
    stream_for Image.find(params[:image_id])
  end
end
```

```ruby
class ImageReflex < StimulusReflex::Reflex
  def like
    image = Image.find(element.id)
    image.likes.increment
    image.save
    cable_ready['document'].morph({selector: "#image_#{image.id}", html: ApplicationController.renderer.render(partial: "images/image", locals: {image: image}})
    cable_ready.broadcast_to(image)
  end
end
```

I'm doing this without the benefit of a test environment, so assume that this is pseudocode... but hopefully it gets the idea across. You could have an arbitrary number of these images on your page, updating counters in real-time, with basically zero client **code** much less client state. So the whole premise of needing React and Redux to make this viable is absurd.

Please let me know if any of this needs "more".

leastbad:
Just as a heads-up, I woke up today thinking I might have a slight refactor in mind. Stay tuned!

leastbad:
Alright, I made a significant change to the way this PR works and it is far better than yesterday. Ready for prime time!

StephenFiser:
@leastbad what's the reason for calling `@channels[identifier].send(:broadcast_to, model)` as opposed to just `@channels[identifier].broadcast_to(model)`?

StephenFiser:
Also - would it make sense to do this?
```ruby
def broadcast_to(model, identifiers = [])
  receiving(identifiers).reject { |identifier| identifier.is_a?(String) }.each do |identifier|
    @channels[identifier].broadcast_to(model)
  end
end

def receiving(identifiers)
   identifiers.any? ? identifiers : channel_identifiers
end

def channel_identifiers
  @channels.values.map(&:identifier)
end
```
This would allow you to broadcast to a one identifier or a subset of your identifiers as opposed to just broadcasting to either one or all. It also guards against strings accidentally being passed in via the identifiers array.

leastbad:
@StephenFiser you are absolutely right about not needing send. That was just a vestigal holdover from a long-gone iteration.

I'm probably too sleepy to properly parse what you're going for in your second comment. It's not clear if you imagine this being just for `broadcast_to` or `broadcast` as well, for example.

If you consult what's currently on master - which is largely unchanged since the project launched - you'll see that I'm actually giving folks some flexibility that doesn't currently exist in the library. I'm not going to be changing the API of the existing `broadcast` method, which is already in production on many apps. Adding an optional parameter is as radical as I can be. Given that I want to keep the API for `broadcast_to` as similar to `broadcast` as possible, it's not likely that this new mechanism will diverge far from where it started.

It's also important to say that I've never actually seen a scenario where more than one identifier is in play at once in the wild, ever, period. If I was designing the library from scratch today, I wouldn't have allowed it. So while I like the idea of giving people limitless flexibility, this PR already dramatically expands the flexibility of the library far beyond what's possible today without breaking compatibility or frankly, giving people another thing that they feel they have to learn or think about.

So yeah, I'm not yet convinced on the second point. Still listening, though. I just suspect most folks will be calling with 0 or 1 parameter (for resource_to) 95+ of the time.

lib/cable_ready/channel.rb
```diff
@@ -131,7 +131,14 @@ def clear
...
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
-      ActionCable.server.broadcast name, "cableReady" => true, "operations" => operations
+      ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
+      clear
+    end
+
+    def broadcast_to(model)
+      operations.select! { |_, list| list.present? }
+      operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
+      identifier.broadcast_to model, "cableReady" => true, "operations" => operations
```
n-rodriguez, line 141: This doesn't work. It throws an error :

```sh
undefined method `broadcast_to' for "DataChannel":String
```

lib/cable_ready/channels.rb
```diff
@@ -8,16 +8,31 @@ def initialize
...
+        @channels.values.select { |channel| channel.identifier.is_a?(String) }.each(&:broadcast)
+      end
+      clear
+    end
+
+    def broadcast_to(model, identifier = nil)
+      if identifier
+        @channels[identifier].broadcast_to model
+      else
+        @channels.values.reject { |channel| channel.identifier.is_a?(String) }.each do |channel|
```
n-rodriguez, line 32: I think it should be a `select` here

lib/cable_ready/channel.rb
n-rodriguez: call to `constantize` is missing

StephenFiser:
@leastbad I just noticed a simple way to make the feature "mathematically complete" in some sense where you can broadcast to one, some, or all channels. Right now in both `broadcast` and `broadcast_to` it looks like it's either one or all. So yes, you'd probably need to do the same for `broadcast`. Just a thought.

lib/cable_ready/channels.rb
StephenFiser: I have the opposite understanding of this. I think it's expecting classes, not strings, which is why you're having problems with your other comment as well.

lib/cable_ready/channel.rb
leastbad: I believe that you're trying to pass a string as the identifier. If you consult the example above, you'll see that this will generate an error.

I am going to make it raise a more literal exception, so that this is less mysterious. In the meantime, please try calling `cable_ready[DataChannel].broadcast_to(resource)` instead of `cable_ready["DataChannel"].broadcast_to(resource)`.

lib/cable_ready/channels.rb
leastbad: @StephenFiser is correct - this is expecting a valid constant. In your case, `DataChannel` with no quotes.

leastbad:
I really like this. Let me see what I can do.

leastbad:
I just ran with two of @StephenFiser's good ideas, and I think the codebase is better for it now.

The whole process flows much more naturally now. Having an optional splat of identifiers allows for flexibility and automatically filters out params of the wrong type.

leastbad:
I just made a commit that adds a trailing named argument called clear, defaulting to true, to both `broadcast` and `broadcast_to`.

Setting clear: false has the effect of not clearing out the channel operations data after a broadcast. This will conserve resources when sending the same broadcast to multiple resources.

I will add information on this to the PR description.

StephenFiser:
@leastbad at first glance this looks really great. Hopefully going to have some time to play with it later tonight!

StephenFiser:
I ran through a quick demo app build out with this branch. I have to say it works amazingly well. I assume at least part of that feeling is that I'm new to cableready, so there's still a bit of *wow* factor with that alone. That said, here's what I did in a nutshell:

1. Set up a basic projects scaffold with name and text
2. Added two channels, one for an individual project and one for a "dashboard" (projects index)
3. Installed cable ready as prescribed in the docs and switched to your branch (@leastbad)
4. Iteratively added this to project.rb (started with just one channel, then went to both)
```ruby
class Project < ApplicationRecord
  after_save do
    # for dashboard
    cable_ready[DashboardChannel].morph({
      selector: ".project-#{id}-name",
      html: "<td class='project-#{id}-name'>#{name}</td>"
    })

    # for project show
    cable_ready[ProjectChannel].morph({
      selector: ".project-name",
      html: "<span class='project-name'>#{name}</span>"
    })

    cable_ready.broadcast_to(self, DashboardChannel, ProjectChannel)
  end
end
```
Quick (unlisted) video showing it in action: https://youtu.be/GToUDbabdtw

leastbad:
That’s rad! I’m thrilled to see this working and jealous that I rarely get to actually play with the tools. 🍿

leastbad:
Huzzah!

PR #39, opened 2020-05-19
Correct method 'outerHtml' to 'outer_html' in documentation

pskarlas:
Typo found in the outerHTML element mutation example 
:)


leastbad:
Nice catch, @pskarlas!

pskarlas:
@leastbad  &#128077;

PR #42, opened 2020-05-28
Feature/broadcast options

StephenFiser:
This change would allow me to be able to do something like:

```ruby
html = ...
cable_ready["room_channel_#{message.room_id}"].insert_adjacent_html(
  selector: '#messages',
  html: html
)
cable_ready["room_channel_#{message.room_id}"].add_option('user_id', message.user_id)
cable_ready.broadcast
```

See this for context: https://youtu.be/p0p9c_WmnLc

Essentially, I'm doing different things to the UI depending on whether or not I'm the one who posted the message (e.g. text appears on left or right side of screen). I need an easy way to access a user ID for conditional logic even if all of the DOM updates come through CableReady.

### DIFF
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -111,10 +111,10 @@ module CableReady
     #     value:    "string"
     #   }, ...],
     # }
-    def initialize(name)
+    def initialize(name, options = {})
       @name = name
       @operations = stub
-      @options = {}
+      @options = options
     end
 
     def clear
@@ -190,10 +190,6 @@ module CableReady
 
     def set_dataset_property(options = {})
       add_operation(:set_dataset_property, options)
-    end
-
-    def add_option(key, value)
-      @options[key] = value
     end
 
     private
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -120,10 +120,10 @@ module CableReady
       @operations = stub
     end
 
-    def broadcast
+    def broadcast(options = {})
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
-      ActionCable.server.broadcast name, "cableReady" => true, "operations" => operations
+      ActionCable.server.broadcast name, "cableReady" => true, "operations" => operations, "options" => options
       clear
     end
 
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -111,10 +111,10 @@ module CableReady
     #     value:    "string"
     #   }, ...],
     # }
-    def initialize(name, options = {})
+    def initialize(name)
       @name = name
       @operations = stub
-      @options = options
+      @options = {}
     end
 
     def clear
@@ -190,6 +190,10 @@ module CableReady
 
     def set_dataset_property(options = {})
       add_operation(:set_dataset_property, options)
+    end
+
+    def add_option(key, value)
+      @options[key] = value
     end
 
     private
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -1,6 +1,6 @@
 module CableReady
   class Channel
-    attr_reader :name, :operations
+    attr_reader :name, :operations, :options
 
     # Example Operations Payload:
     #
@@ -111,16 +111,17 @@ module CableReady
     #     value:    "string"
     #   }, ...],
     # }
-    def initialize(name)
+    def initialize(name, options = {})
       @name = name
       @operations = stub
+      @options = options
     end
 
     def clear
       @operations = stub
     end
 
-    def broadcast(options = {})
+    def broadcast
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
       ActionCable.server.broadcast name, "cableReady" => true, "operations" => operations, "options" => options
--- a/lib/cable_ready/channels.rb
+++ b/lib/cable_ready/channels.rb
@@ -6,8 +6,8 @@ module CableReady
       @channels = {}
     end
 
-    def [](channel_name)
-      @channels[channel_name] ||= CableReady::Channel.new(channel_name)
+    def [](channel_name, options = {})
+      @channels[channel_name] ||= CableReady::Channel.new(channel_name, options)
     end
 
     def clear
--- a/lib/cable_ready/channels.rb
+++ b/lib/cable_ready/channels.rb
@@ -6,8 +6,8 @@ module CableReady
       @channels = {}
     end
 
-    def [](channel_name, options = {})
-      @channels[channel_name] ||= CableReady::Channel.new(channel_name, options)
+    def [](channel_name)
+      @channels[channel_name] ||= CableReady::Channel.new(channel_name)
     end
 
     def clear
--- a/lib/cable_ready/version.rb
+++ b/lib/cable_ready/version.rb
@@ -1,3 +1,3 @@
 module CableReady
-  VERSION = "4.1.3"
+  VERSION = "4.1.4"
 end
--- a/lib/cable_ready/version.rb
+++ b/lib/cable_ready/version.rb
@@ -1,3 +1,3 @@
 module CableReady
-  VERSION = "4.1.4"
+  VERSION = "4.1.3"
 end
--- a/lib/cable_ready/version.rb
+++ b/lib/cable_ready/version.rb
@@ -1,3 +1,3 @@
 module CableReady
-  VERSION = "4.1.2"
+  VERSION = "4.1.3"
 end


hopsoft:
Thanks for this PR. I want to watch your first video again to better grok what you bumped into that led to this contribution. Will review & test soon.

leastbad:
Hey @StephenFiser! I really enjoyed your video and wanted to offer some ideas that would get you where you want to be.

Before I forget: the default position for insert_adjacent_html is actually `beforeend` so you can skip that if you want.

I'm a bit confused as to why you're seeing a flash when the insert occurs; it's almost like it attaches to the DOM, does a draw call and then realizes that it's support to inherit CSS styles from whatever class you have applied. This is something we should look into further; in the meantime, you could experiment with throwing in an opacity transition animation on those message elements, just something to keep it invisible for 17ms while it figures out what to draw on the next "frame".

At 13:08 you said "that is quite strange" and I LOL'd because it wasn't strange the first time you looked at the object - you only said that once you had packed your `user_id` into the object! Given how brutal integrations with new libraries can be, I'd describe the ease at which you were able to build on CR's defaults as "quite awesome". 😈 

Anyhow, I have some pretty radical suggestions that will require you to rethink how you structure your AC channels and probably your DOM ids, but they are in the spirit of how both ActionCable and CableReady are supposed to work, so best to suffer the refactor now and feel like a superhero the next time it comes up. I should mention too that not knowing about what I'm about to explain is not your fault - the ActionCable documentation is an embarrassment.

The ActionCable guide describes how to set up your `room_channel.rb` which I can't see in this video, but the general idea is that you have a `subscribed` method and in that method, you `stream_from "room_channel"` or something. And yes, of course that `"room_channel"` could be `"room_#{params[:room]}_channel"` or something. But at the end of the day, it's just a way of specifying a token that you can subscribe to. Information is broadcasting over the RoomChannel, but that channel can have an unlimited number of subscription tokens. This is, charitably, confusing AF and never really broken down. You can stream_from `"room_channel"` or `"room_1_channel"` or `"clownfart"` and it's all going to be consumed by people consuming the Room channel. If you're not confused, that's **great** because it's like they *want* you to be confused. And that's frustrating because it really is such an amazing library.

The part that they don't tell you is that there's a different way to stream broadcasts, and it's almost certainly what you're looking for? What if I told you that instead of streaming_from a string, you could stream_for a resource?

```ruby
class RoomChannel < ApplicationCable::Channel
  def subscribed
    stream_for User.find(params[:id])
  end
end
```

woah

For example, I have a polymorphic comment class that you can attach to any other model:

```ruby
class CommentChannel < ApplicationCable::Channel
  def subscribed
    stream_for params[:commentable_type].constantize.find(params[:commentable_id])
  end
end
```
and the relevant snippet from my client code:
```js
consumer.subscriptions.create(
  {
    channel: 'CommentChannel',
    commentable_type: 'Post',
    commentable_id: 1
  },
  {
    received (data) {
      // do cool stuff
    }
  }
```

The point is that you can subscribe to your room as a user, and receive updates based on the user. So it's less about filtering out the wrong data and giving AC the information it needs to only send it to the right clients.

To support this, there's a PR #38 that will introduce a `broadcast_for` method to CableReady, hopefully **coming soon**.

However, I believe that you can do what you need to do today, without adding an options parameter to the library - which, if you're reading between the lines, isn't terribly likely to go through at this point - by rethinking how you approach setting up your subscription strings with `stream_from`. It will require a modest refactoring but it's mostly about shifting how your brain approaches this kind of problem going forward. The investment will be worth it, and `broadcast_for` will just be gravy.

StephenFiser:
@hopsoft thank you so much for the thorough reply! I will try to implement your ideas and see where I end up. CR was actually quite easy to get started with. It's more a problem on the actioncable side wrt the docs (as you mentioned).

I actually started this series on actioncable to try to publicly work through the oddities and see what I could figure out. Glad you liked the video!

StephenFiser:
@hopsoft & @leastbad - I've read through this a few times and here's one thing I'm still not understanding. Right now, I essentially have users subscribed to a room, so my channel just looks like:

```ruby
class RoomChannel < ApplicationCable::Channel
  def subscribed
    stream_from "room_channel_#{params[:room_id]}"
  end
end
```
I get that I could switch this to something like:
```ruby
class RoomChannel < ApplicationCable::Channel
  def subscribed
    stream_for Room.find(params[:room_id])
  end
end
```
but I'm not sure that helps with the issue I'm having. (I'm really happy to be wrong, but right now I don't quite get it.)

Let me see if I can better explain the use case.

I've essentially got a demo app (in the video) that looks a lot like iOS messenger or FB messenger. Messages that I write appear on the right side of the screen in blue, and other people's messages are on the left side in gray.

The controller for creating a message kicks off a job that broadcasts the messages to everyone subscribed to the channel and then redirects.

Looks like this:
```ruby
def create
  @message = current_user.messages.create(message_params)
  SendMessageJob.perform_later(@message)
  redirect_to @message.room
end
```

The reason for the "flicker" where the message jumps from the left side of the screen to the right is because the message briefly gets added to the page on the left by via actioncable before the user's screen is refreshed (via turbolinks) from the redirection.

My solution was to just not append the message to the window via actioncable for the user who posted the message - just let the page reload handle appending it.

To do that, I have something like this:
```html
<div id='current-user-id' data-id='<%= current_user.id %>'></div>
```

Then when I get into the client side from a broadcast, I'm checking whether the `user_id` of the newly posted message matches that of the `current_user` in the browser.

As far as I understand it, I either need to be able to:
a) broadcast from the server to everyone *except* the user who posted the message or
b) do something similar to my solution and filter out / display the message differently based on whether the `user_id` in the browser matches the `message.user_id` (or similar)

StephenFiser:
Last comment for now 😬😅

Suppose I were to do something like you suggested @leastbad :

```ruby
class RoomChannel < ApplicationCable::Channel
  def subscribed
    stream_for User.find(params[:id])
  end
end
```
They have this in the docs:
```ruby
class CommentsChannel < ApplicationCable::Channel
  def subscribed
    post = Post.find(params[:id])
    stream_for post
  end
end

CommentsChannel.broadcast_to(@post, @comment)
```
This would seem to imply that I need to keep track of which users are in which room at all times and then do something like:
```ruby
room.users.each do |user|
  RoomChannel.broadcast_to(user, message)
end
```
Does that seem right?

If that works, that would allow me to do something like:
```ruby
room.users.each do |user|
  RoomChannel.broadcast_to(user, message) unless message.sent_by?(user)
end
```

If so, how would this work in CableReady? Right now I have this in my `SendMessageJob`:
```ruby
cable_ready["room_channel_#{message.room_id}"].insert_adjacent_html(
  selector: '#messages',
  html: html
)
cable_ready.broadcast
```
Is this basically where `broadcast_for` is going to come into play?

Apologies for all of the questions. Just trying to get my head around it.

leastbad:
Hey @StephenFiser! While I can't build your app for you, I actually love engaging on this kind of design discussion. You should join the StimulusReflex [Discord](https://discord.gg/XveN625) if you haven't already - there's a hang of smart folks that will dig in on these sorts of challenges for fun pretty much 24/7. You'll find a #cableready channel there.

We're all huge Turbolinks fans, but I would not use Turbolinks to refresh the page as you do here. I would instead either use StimulusReflex or an Ajax/fetch call to post the new message to the server, and let CableReady append it to the conversation without a refresh.

I'm not sure if you're using Stimulus. You should be! You are already using both Turbolinks and CableReady. Stimulus is awesome because everything it does is based on watching your DOM with a MutationObserver. So lets say that you append a message with CableReady, and the DIV containing your message has a `data-controller="message"` attribute on it, it will run that controller instance's `connect()` callback when the DIV is appended. This is extraordinarily powerful and was explicitly designed to work very well with Turbolinks.

You absolutely could use a Stimulus controller to add your new message to the right side of the discussion UI, while relying on the server to send the new message to everyone but the sender.

As for the bigger picture solution, regardless of whether you use stream_from or stream_for, I would be strongly tempted to think of your subscription identier as being keyed to both the room/discussion and the participant, so in reality each person is subscribing to their own version of the feed. It's not so different from the often long cache keys you end up with when doing Russian doll caching, because you need every variable to be accounted for.

This could be as "simple" as `stream_from "room_#{params[:room_id]}_user_#{params[:user_id]}"` or you could experiment with doing `stream_for Room.find(params[:id])` and yes... loop through the participants and broadcast to everyone except the sender. They should already see their own message thanks to whatever you did on the client, ideally with Stimulus.

Again, I can't write your app for you but you're on the right track when you think of every user having a subscription to a feed that is customized for them. It is also via this thinking that allows you to send messages to individual users that others can't see, as when Slack welcomes people to channels or asks them if they want to post a Dropbox preview or just the link.

StephenFiser:
Thanks @leastbad! That totally clears things up. Going to go ahead and close this PR. Just FYI - This app is completely for education purposes on Youtube. I've recorded 100% of the dev work and published. Once I get my head around ActionCable, CableReady, StimulusReflex, I'm going to be putting out a lot of video tutorials and how-to content. I've been doing content for a while, but I got particularly annoyed with the lack of info on actioncable, so I decided to start making some.

leastbad:
Great stuff, Stephen. Please keep us in the loop, especially if you're making videos. If they cover interesting ground, we'll almost certainly end up promoting and linking to them.

We're also actively discussing launching some community resources for both Stimulus as a standalone library and separately, this new technology stack we're iterating on and innovating with... roughly Rails + StimulusReflex + Turbolinks + Stimulus + CableReady, with supporting players like Russian doll caching, Sidekiq and Redis along for the ride. We believe that these tools represent a major sea change in how UIs will be built going forward, hopefully presenting a compelling alternative to the client-side insanity that follows tools like React. It'd be great to have you contributing to these efforts.

StephenFiser:
That's awesome. Would love to discuss once I get my head around things a bit better! Right now the videos I'm making are just code-alongs where I try to sort things out. Once I get a better grip, I'll be doing more step-by-step how-to videos.

leastbad:
Just getting to the end of the video. Thanks for the shout-out. :)

You're definitely getting there. You might find some value in keeping the `if (data.cableReady)` just to make sure that you're performing the operations you think you are. If something else ever came down the pipe, you down want CableReady exploding by accident.

The other thing is that you can actually move the call to `cable_ready.broadcast` out of the .each block. When you call `insert_adjacent_html` or any of the other operations, you're actually putting an instruction into an array of operations that will be fired.

You can prime `cable_ready` with a list of operations and call `broadcast` to send them all at once.

PR #46, opened 2020-06-01
after-outer-html emitted from new element

leastbad:


javascript/cable_ready.js
```diff
@@ -150,9 +150,14 @@ const DOMOperations = {
     activeElement = document.activeElement
     const { element, html, focusSelector } = detail
     dispatch(element, 'cable-ready:before-outer-html', detail)
+    const ordinal = Array.from(element.parentElement.children).indexOf(element)
```
julianrubisch, line 153:
```suggestion
    const parent = element.parentElement
    const ordinal = Array.from(parent.children).indexOf(element)
```

javascript/cable_ready.js
```diff
@@ -150,9 +150,14 @@ const DOMOperations = {
...
     element.outerHTML = html
     assignFocus(focusSelector)
-    dispatch(element, 'cable-ready:after-outer-html', detail)
+    dispatch(
+      element.parentElement.children[ordinal],
```
julianrubisch, line 157:
```suggestion
      parent.children[ordinal],
```

Issue #43, opened 2020-05-31
Make JS DOMOperations extendable


n-rodriguez:
Hi there! First, thank you for this nice gem :)

Second, I think it would be nice to make `DOMOperations` extendable to build our own feature on CableReady.

Ruby is already extendable with a bit of monkey patching : 

```ruby
# frozen_string_literal: true

require 'cable_ready/channel'

# Patch CableReady::Channel to add our own operations

module Concerto
  module CoreExt
    module CableReadyPatch

      def notify(options = {})
        add_operation(:notify, options)
      end


      def stub
        super.merge(notify: [])
      end

    end
  end
end

CableReady::Channel.prepend(Concerto::CoreExt::CableReadyPatch)
```
For JS it's doable but it needs a patch :

The patch :

```js
export default { perform, isTextInput, DOMOperations }
```

Then in application :

```coffee
import CableReady  from 'cable_ready/cable_ready_patched'

CableReady.DOMOperations['notify'] = (detail) =>
  AsyncEvents.display_growl_message(detail)
```

```ruby
cable_ready['MyChannel'].notify(opts)
cable_ready.broadcast_to(user.private_channel, 'MyChannel')
```

Note: the example make use of this PR : https://github.com/hopsoft/cable_ready/pull/38

Thank you!

n-rodriguez:
Thank you!

leastbad:
It's a small change but a really cool feature. Thanks @n-rodriguez!

PR #40, opened 2020-05-22
syntax fix : outerHTML to outer_html

JeromeSadou:

PR #47, opened 2020-06-07
Customize operations via initializer

leastbad:
I spent a lot of time looking at @n-rodriguez' #43 because I **hate** that he has to monkey patch the library in order to add additional operations. There's been a few moments where we've idly talked about supporting configuration details and also I have long been irritated by the implicit singleton pattern implied in the way we use CableReady::Channels. It turns out that Ruby has a workable singleton module that we can use to start refacting how CR works behind the scenes, allowing for customization.

This PR would allow @n-rodriguez to just create `config/initializers/cable_ready.rb`:

```ruby
CableReady::Channels.configure do |config|
  config.add_operation :notify
end
```

Thanks to @foca, @n-rodriguez also has the option to define the implementation of their new custom operation:

```ruby
CableReady::Channels.configure do |config|
  config.add_operation :notify do |options|
    puts "Eat dirt!"
  end
end
```

Having a list of operations means that the code for `channel.rb` is dramatically more concise, as we can dynamically define everything in a consistent way. Adding a new operation to the library is now just adding a symbol to an array.

This also opens the door to additional future configuration opportunities via the initializer.

All of the function signature comments have been erased as we now have proper project documentation, keeping everything in one place and drastically lowering both tedium and the incident rate of typing errors, as we saw recently with #40.

Note that there is an extremely minor *technically* breaking change which I suspect will only impact CodeFund as it's not even documented, yet: for reasons that are unknown to me (but I assume good intentions) all of the operation method signatures were identical `options = {}` **except for set_cookie** which inexplicably accepted a value string. In order for this to work, anyone using set_cooking would have to update from "string" to {cookie: "string"}. **Technically, we could define a custom implementation for `set_cookie` that would keep the current function signature, but my vote is to move forward with a consistent API for all "factory settings".**

### DIFF
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -8,14 +8,14 @@ module CableReady
       @name = name
       @tools = tools
       @operations = stub
-      
+
       tools.each do |tool|
         define_singleton_method tool do |options = {}|
           add_operation tool, options
         end
       end
     end
-    
+
     def broadcast
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
@@ -31,9 +31,8 @@ module CableReady
     end
 
     def stub
-      tools.inject({}) do |hash, tool|
+      tools.each_with_object({}) do |tool, hash|
         hash[tool] = []
-        hash
       end
     end
   end
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -2,209 +2,25 @@
 
 module CableReady
   class Channel
-    attr_reader :name, :operations
+    attr_reader :name, :operations, :tools
 
-    # Example Operations Payload:
-    #
-    # {
-    #   # Cookies .....................................................................................................
-    #
-    #   set_cookie: [
-    #     {cookie: "example=value; path=/; expires=Sat, 07 Mar 2020 16:19:19 GMT"},
-    #     ...
-    #   ]
-    #
-    #   # DOM Events ..................................................................................................
-    #
-    #   dispatch_event: [{
-    #     name:     "string",
-    #     detail:   "object",
-    #     selector: "string",
-    #   }, ...],
-    #
-    #   # Element Mutations ...........................................................................................
-    #
-    #   morph: [{
-    #     selector:      "string",
-    #     html:          "string"
-    #     children_only:  true|false,
-    #     permanent_attribute_name: "string",
-    #     focus_selector: "string",
-    #   }, ...],
-    #
-    #   inner_html: [{
-    #     selector:      "string",
-    #     focus_selector: "string",
-    #     html:          "string"
-    #   }, ...],
-    #
-    #   outer_html: [{
-    #     selector:      "string",
-    #     focus_selector: "string",
-    #     html:          "string"
-    #   }, ...],
-    #
-    #   text_content: [{
-    #     selector: "string",
-    #     text:     "string"
-    #   }, ...]
-    #
-    #   insert_adjacent_html: [{
-    #     selector:      "string",
-    #     focus_selector: "string",
-    #     position:      "string",
-    #     html:          "string"
-    #   }, ...],
-    #
-    #   insert_adjacent_text: [{
-    #     selector: "string",
-    #     position: "string",
-    #     text:     "string"
-    #   }, ...],
-    #
-    #   remove: [{
-    #     selector:      "string",
-    #     focus_selector: "string,
-    #   }, ...],
-    #
-    #   set_property: [{
-    #     name:     "string",
-    #     selector: "string",
-    #     value:    "string"
-    #   }, ...],
-    #
-    #   set_value: [{
-    #     selector: "string",
-    #     value:    "string"
-    #   }, ...],
-    #
-    #   # Attribute Mutations .........................................................................................
-    #
-    #   set_attribute: [{
-    #     selector: "string",
-    #     name:     "string",
-    #     value:    "string"
-    #   }, ...],
-    #
-    #   remove_attribute: [{
-    #     selector: "string",
-    #     name:     "string"
-    #   }, ...],
-    #
-    #   # CSS Class Mutations .........................................................................................
-    #
-    #   add_css_class: [{
-    #     selector: "string",
-    #     name:     "string"
-    #   }, ...],
-    #
-    #   remove_css_class: [{
-    #     selector: "string",
-    #     name:     "string"
-    #   }, ...],
-    #
-    #   # Style Mutations ...........................................................................................
-    #
-    #   set_style: [{
-    #     selector: "string",
-    #     name:     "string",
-    #     value:    "string"
-    #   }, ...],
-    #
-    #   # Dataset Mutations ...........................................................................................
-    #
-    #   set_dataset_property: [{
-    #     selector: "string",
-    #     name:     "string",
-    #     value:    "string"
-    #   }, ...],
-    # }
-    def initialize(name)
+    def initialize(name, tools)
       @name = name
+      @tools = tools
       @operations = stub
+      
+      tools.each do |tool|
+        define_singleton_method tool do |options = {}|
+          add_operation tool, options
+        end
+      end
     end
-
-    def clear
-      @operations = stub
-    end
-
+    
     def broadcast
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
       ActionCable.server.broadcast name, "cableReady" => true, "operations" => operations
-      clear
-    end
-
-    def set_cookie(value)
-      add_operation(:set_cookie, {cookie: value})
-    end
-
-    def dispatch_event(options = {})
-      add_operation(:dispatch_event, options)
-    end
-
-    def morph(options = {})
-      add_operation(:morph, options)
-    end
-
-    def inner_html(options = {})
-      add_operation(:inner_html, options)
-    end
-
-    def outer_html(options = {})
-      add_operation(:outer_html, options)
-    end
-
-    def text_content(options = {})
-      add_operation(:text_content, options)
-    end
-
-    def insert_adjacent_html(options = {})
-      add_operation(:insert_adjacent_html, options)
-    end
-
-    def insert_adjacent_text(options = {})
-      add_operation(:insert_adjacent_text, options)
-    end
-
-    def remove(options = {})
-      add_operation(:remove, options)
-    end
-
-    def set_property(options = {})
-      add_operation(:set_property, options)
-    end
-
-    def set_value(options = {})
-      add_operation(:set_value, options)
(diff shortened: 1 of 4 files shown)


lib/cable_ready/channels.rb
```diff
@@ -4,12 +4,24 @@
...
+    end
+
     def initialize
       @channels = {}
+      @tools = [:add_css_class, :dispatch_event, :inner_html, :insert_adjacent_html, :insert_adjacent_text, :morph, :outer_html, :remove, :remove_attribute, :remove_css_class, :set_attribute, :set_cookie, :set_dataset_property, :set_property, :set_style, :set_styles, :set_value]
+    end
+
+    def add_tool(tool)
+      @tools << tool.to_sym
     end
```
foca, line 21: Would it make sense to allow a block here that provides the body of the tool's method?

Something like

``` ruby
def add_tool(tool, &implementation)
  @tools[tool] = implementation || ->(options = {}) { add_operation(tool, options) }
end
```

And then

``` ruby
def initialize(name, tools)
  ...

  tools.each do |tool, implementation|
    define_singleton_method(tool, &implementation)
  end
end
```

I don't really have a specific use case in mind, and this might be making internal APIs more public than they need to be, but if the goal is to reduce the necessity for monkey patching that this could come in handy if someone needs to do something more custom in their operation.

WDYT?

leastbad: I *love* this idea. I'm just going to borrow it and pretend like it was my idea, okay? ;)

Seriously - this is great.

lib/cable_ready/channel.rb
```diff
@@ -2,209 +2,25 @@
...
     def broadcast
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
       ActionCable.server.broadcast name, "cableReady" => true, "operations" => operations
-      clear
-    end
-
-    def set_cookie(value)
-      add_operation(:set_cookie, {cookie: value})
-    end
```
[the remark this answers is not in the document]
leastbad: Serious question: did you read the last paragraph of the PR?

I don't mean that in a shitty way... I'm trying to understand if you missed what I said, or if what I said doesn't cover it and I'm missing something.

Quick recap: this is a minor breaking change but one I'm currently advocating for because it's not even documented with and the number of people currently using it could be 0 or 1 today and I'm not sure why it even had to be different in the first place. I **strongly** suspect that it was an oversight.

leastbad:
Great feedback re: tools

I did attempt to explain in the PR that conceptually, tools are what's possible vs operations are what's queued up.

The main reason I introduced an @tools object is that there is already @operations... which, again, is the list of operations that are queued up. Before this PR, there was no explicit concept of features-as-atomic-elements. My first instinct was to call them operations, but that namespace was taken - both by the code itself and by the current implementation where you CableReady.perform(data.operations).

I could have called it @features but I went with @tools. I'm super-duper open to changing it to something better, but then we have to sort out what happens with the existing @operations.

lib/cable_ready/channel.rb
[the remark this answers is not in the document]
foca: > Quick recap: this is a minor breaking change but one I'm currently advocating for because it's not even documented with and the number of people currently using it could be 0 or 1 today and I'm not sure why it even had to be different in the first place. I **strongly** suspect that it was an oversight.

DOH, sorry, I missed that part of the text. I should probably read the whole thing before I start switching before text and code 😬

foca:
Sorry, I might've not explained myself properly. `@operations` is taken inside `Channel`, but not in the `Channels` singleton, which with this PR, is becoming a configuration object. In that sense, you'd end up with this API

``` ruby
Channels.operations #=> list of available operations
Channels.add_operation(op, &implementation) #=> add a new possible operation
```

and then

``` ruby
Channel#operations #=> queue of things to do in the front-end after send
```

Since the list of tools/ops passed from `Channels` to `Channel` is only passed to the initializer and "dies" there as the entries become methods, there wouldn't be any naming clashes happening.

---

That said, I can see how overloading the term could be confusing, so I understand if a new name is better. I mostly see it as referring to the same thing at different levels of abstraction (configuring a thing vs doing a thing), so in my mind I'd call them the same.

leastbad:
@foca ... sort of like that? 😁

lib/cable_ready/channel.rb
```diff
@@ -215,25 +29,9 @@ def add_operation(key, options)
...
-        set_attribute: [],
-        set_cookie: [],
-        set_dataset_property: [],
-        set_style: [],
-        set_property: [],
-        set_value: [],
-        text_content: []
-      }
+      available_operations.each_with_object({}) do |available_operation, hash|
+        hash[available_operations] = []
```
hopsoft, line 33: You have a typo here `available_operations` should be `available_operation`.

hopsoft:
This is a nice evolutionary step for the library. I'm looking forward to testing it out.

lib/cable_ready/channel.rb
leastbad: lol nice catch

leastbad:
I realized all of a sudden that I had changed the server side method signature for set_cookie, but didn't even look at the client JS. Shame on me!

In addition to adjusting setCookie() to use the information provided by the server, I added event dispatches for before and after-set-cookie.

I did a search on GitHub to see if anyone is using set_cookie today, and it doesn't seem like it: https://github.com/search?q=CableReady+set_cookie&type=Code

While the original PR#28 identified set_cookie as a PoC, let's have the talk about whether we should do this the dumb way (as it is now, where it accepts an arbitrary non-interpolated string) or if we should attempt to provide a simple API for building a cookie. We could also do **both**, leaving the current dumb string support but also adding the ability to build it up programmatically. The real question is: *how much of a use case is this, really*? Note: when I say dumb, I'm referring to the not-validated cookie setting string, not the developer. Dumb as in hammer.

I actually just saw that @jonathan-s [arrived at the same conclusion in April](https://github.com/hopsoft/cable_ready/pull/28#issuecomment-620588270). Because of course he did. 😀

lib/cable_ready/channel.rb
```diff
@@ -215,25 +29,9 @@ def add_operation(key, options)
...
-        remove_css_class: [],
-        set_attribute: [],
-        set_cookie: [],
-        set_dataset_property: [],
-        set_style: [],
-        set_property: [],
-        set_value: [],
-        text_content: []
-      }
+      available_operations.each_with_object({}) do |available_operation, hash|
```
dlt, line 32: I'm not 100% sure, but it seems that we're reimplementing std lib functionality here. I think that the `@operations` variable could be initialized with a hash that accepts a block, instead of receiving the return of `#stub`, like:

```ruby
# in the initializer
@operations = Hash.new { |hash, operation_name| hash[operation_name] = [] }
```

lib/cable_ready/channels.rb
```diff
@@ -4,12 +4,27 @@
...
+    attr_accessor :operations
+
+    def self.configure
+      yield CableReady::Channels.instance if block_given?
+    end
+
     def initialize
       @channels = {}
+      @operations = {}
+      [:add_css_class, :dispatch_event, :inner_html, :insert_adjacent_html, :insert_adjacent_text, :morph, :outer_html, :remove, :remove_attribute, :remove_css_class, :set_attribute, :set_cookie, :set_dataset_property, :set_property, :set_style, :set_styles, :set_value].each do |operation|
```
dlt, line 17: Have you considered using the literal notation for an array of symbols? It would avoid `,` and `:` and make the code a little bit easier to read.

leastbad: Do you mean %w[ this burns my eyes ]

I hadn't considered it, because that generates an array of strings, not Symbols.

Honestly, I have never loved %w[] and it could just be subjective, but I feel like an array of Symbols is more idiomatic in the wild. What do you think, @hopsoft?

lib/cable_ready/channel.rb
leastbad: Wow, nice call. I was pretty confident that you must be missing something, but it's working great! 😁 

It's pretty crazy how refactoring something really verbose suddenly reveals ways in which things can get even more streamlined.

leastbad:
I brought in the latest from master, so `broadcast_to` is now present and accounted for.

Really just paused for @hopsoft to find some time to give this a spin and offer an opinion on whether we should switch https://github.com/hopsoft/cable_ready/pull/47/files#diff-f561555d3426cd98514ec3f00e6cf569R17 to use %w or %i array notation vs leaving it as it is, given we don't use literal notation elsewhere in CR or SR. I will support whatever decision is reached 100%.

lib/cable_ready/channels.rb
hopsoft: Here's my preferred syntax/formatting for lists like this.

```ruby
operation_names = %i[
  add_css_class
  dispatch_event
  inner_html
  insert_adjacent_html
  insert_adjacent_text
  morph
  outer_html
  remove
  remove_attribute
  remove_css_class
  set_attribute
  set_cookie
  set_dataset_property
  set_property
  set_style
  set_styles
  set_value
]
```

lib/cable_ready/channel.rb
```diff
@@ -2,245 +2,35 @@
...
+        define_singleton_method available_operation, &implementation
+      end
     end
 
     def broadcast(clear)
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
       ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
-      reset if clear
+      @operations = Hash.new { |hash, operation| hash[operation] = [] } if clear
```
n-rodriguez, line 20: Maybe extract this in its own private method `reset_operations` or `clear_operations`

lib/cable_ready/channel.rb
```diff
@@ -2,245 +2,35 @@
...
-      add_operation(:set_style, options)
-    end
-
-    def set_styles(options = {})
-      add_operation(:set_styles, options)
-    end
-
-    def set_dataset_property(options = {})
-      add_operation(:set_dataset_property, options)
+      @operations = Hash.new { |hash, operation| hash[operation] = [] } if clear
```
n-rodriguez, line 27: Same as above

lib/cable_ready/channel.rb
```diff
@@ -2,245 +2,35 @@
...
-    def initialize(identifier)
+    def initialize(identifier, available_operations)
       @identifier = identifier
-      @operations = stub
-    end
-
-    def reset
-      @operations = stub
+      @available_operations = available_operations
+      @operations = Hash.new { |hash, operation| hash[operation] = [] }
```
n-rodriguez, line 10: Maybe extract this in its own private method reset_operations or clear_operations

leastbad: Good call.

leastbad:
🥳

n-rodriguez:
Thank you!

PR #58, opened 2020-07-06
only pass inner html if childrenOnly is false

joshleblanc:
Okay, so this is the result of a lot of iteration trying to get `children_only: false` to work properly with `data-reflex-permanent`. If we just keep `template.content` when using `children_only: false`, you end up with a number of strange situations, dending on the HTML you're trying to morph. I believe this is due to the differences between fromNode and toNode. This change results in my test project successfully rendering with children_only set to true or false, always respecting the permanent attribute.

I'm not 100% confident about the ramifications of this change, but I'm testing it in my expo app and everything's working as expected for all examples. 

This is related to issue 2 on the discord.

> For reasons we don't currently understand, when you do a :selector morph, it ignores data-reflex-permanent attributes. (If you're familiar with cable_ready.js, shouldMorph just does not fire.) This is confounding and very frustrating. It seems to happen regardless of whether children are replaced.

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -160,8 +160,7 @@ const DOMOperations = {
     })
     const parent = element.parentElement
     const ordinal = Array.from(parent.children).indexOf(element)
-    const toNode = childrenOnly ? template.content : template.innerHTML
-    morphdom(element, toNode, {
+    morphdom(element, childrenOnly ? template.content : template.innerHTML, {
       childrenOnly: !!childrenOnly,
       onBeforeElUpdated: shouldMorph(permanentAttributeName)
     })
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -160,7 +160,8 @@ const DOMOperations = {
     })
     const parent = element.parentElement
     const ordinal = Array.from(parent.children).indexOf(element)
-    morphdom(element, template.content, {
+    const toNode = childrenOnly ? template.content : template.innerHTML
+    morphdom(element, toNode, {
       childrenOnly: !!childrenOnly,
       onBeforeElUpdated: shouldMorph(permanentAttributeName)
     })


leastbad:
I'm intrigued by this and it's possibly the first time in my career as a Rails developer I've wished that I could have access to a server side DOM so that we wouldn't have to consider this change, as it would be much easier to just pass the right HTML.

Unfortunately, as smart as this change seems to be, it is a breaking change. Maybe it's the right thing to do! But someone relying on this in the wild who expects the old behaviour is going to be in for a rude surprise.

Anyhow, let's assume (wrongly) that StimulusReflex is the only project using this function and we can make this change... this would make it impossible to have morphdom update the container if you wanted it to, wouldn't it?

joshleblanc:
It still replaces the selected element as it should when children_only is false.

QN8wKn6wDd

hopsoft:
Thanks for this @joshleblanc I'll be testing this week and very likely getting this merged soon.

javascript/cable_ready.js
```diff
@@ -160,7 +160,7 @@ const DOMOperations = {
     })
     const parent = element.parentElement
     const ordinal = Array.from(parent.children).indexOf(element)
-    morphdom(element, template.content, {
+    morphdom(element, childrenOnly ? template.content : template.innerHTML, {
```
kylefox, line 163: FWIW I _think_ `firstElementChild` will have the same net effect as passing `innerHTML`, but will save morphdom from having to [construct an Element from a string](https://github.com/patrick-steele-idem/morphdom/blob/master/src/morphdom.js#L26-L34):

```suggestion
    morphdom(element, childrenOnly ? template.content : template.content.firstElementChild, {
```

cc @hopsoft

leastbad:
I'm really excited about this change, even if I wish that morphdom would recognize this as a defect. There's a chance we're missing a good reason for morphdom to not play nicely with fragments, but I don't currently see it.

Issue #59, opened 2020-07-11
Event Dispatch: help with serialization


anaice:
Hi guys. Can you help me with a question?
I'm using cable_ready to dispatch a custom DOM event. When I retrieve event.detail it is in camelCase, I would like it to keep my attributes format. Example: `user_id`, not `userId`
```ruby
   def trigger_dom_evt_new_driver_position(user_id, tracking)
     #...
     cable_ready[stream_name].dispatch_event(
        name: evt_id,
        detail:  { user_id: user_id, tracking: TrackingBlueprint.render_as_hash(tracking) },
        selector: "#operational-map-controller"
     )
     cable_ready.broadcast
   end
```
is there any configuration for serialization?

thank you very much

leastbad:
@hopsoft fwiw I agree with @anaice that we shouldn't automatically camelize details. Sadly, fixing this is a breaking minor change.

hopsoft:
I'm a strong believer in the **"When in Rome, do as the Romans do"** philosophy. Most JavaScripters will expect `camelCase` on the JavaScript side while Rubyists expect `snake_case`, but I do like the idea of making it configurable. 

I'd support a PR that makes it configurable that defaults to `camelCase`.

leastbad:
Fair enough. I do think that it's one of the many things that we should explicitly spell out in the documentation. I'll put it on the list!

If @anaice or anyone else is interested in tackling this, please note that we actually do now have support for configuration options via an initializer. Apologies for the awkward lack of real documentation, but you can get the idea from #47.

In short, a `config.details_case` option (or a better name!) is totally doable.

anaice:
Hi guys. I made a pull request for this.
Please forgive any nonsense: old programmer, but new rubyist here.

leastbad:
Rafael, this is amazing news! I haven't even seen the PR yet, but I want to express how proud I am to see you do this.

If it needs work, we'll help.

anaice:
Thanks! Please let me know if I need to improve the PR.

PR #60, opened 2020-07-13
feat(configuration): adds detail_case config

anaice:
#59

Case configuration for detail when operation is `:dispatch_event`. 
Options for `detail_case`:
- camel (default)
- pascal
- snake
- kebab
- keep (keeps original)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,7 @@ PATH
   remote: .
   specs:
     cable_ready (4.3.0)
+      anyway_config (>= 2.0.0)
       rails (>= 5.2)
 
 GEM
@@ -62,6 +63,8 @@ GEM
       minitest (~> 5.1)
       tzinfo (~> 1.1)
       zeitwerk (~> 2.2, >= 2.2.2)
+    anyway_config (2.0.6)
+      ruby-next-core (>= 0.8.0)
     ast (2.4.1)
     builder (3.2.4)
     coderay (1.1.3)
@@ -85,7 +88,7 @@ GEM
     mini_portile2 (2.4.0)
     minitest (5.14.1)
     nio4r (2.5.2)
-    nokogiri (1.10.9)
+    nokogiri (1.10.10)
       mini_portile2 (~> 2.4.0)
     parallel (1.19.2)
     parser (2.7.1.4)
@@ -141,6 +144,7 @@ GEM
       parser (>= 2.7.0.1)
     rubocop-performance (1.6.1)
       rubocop (>= 0.71.0)
+    ruby-next-core (0.9.2)
     ruby-progressbar (1.10.1)
     sprockets (4.0.2)
       concurrent-ruby (~> 1.0)
@@ -159,10 +163,10 @@ GEM
     tzinfo (1.2.7)
       thread_safe (~> 0.1)
     unicode-display_width (1.7.0)
-    websocket-driver (0.7.2)
+    websocket-driver (0.7.3)
       websocket-extensions (>= 0.1.0)
     websocket-extensions (0.1.5)
-    zeitwerk (2.3.0)
+    zeitwerk (2.3.1)
 
 PLATFORMS
   ruby
--- a/cable_ready.gemspec
+++ b/cable_ready.gemspec
@@ -13,6 +13,7 @@ Gem::Specification.new do |gem|
   gem.test_files = Dir["test/**/*.rb"]
 
   gem.add_dependency "rails", ">= 5.2"
+  gem.add_dependency "anyway_config", ">= 2.0.0"
 
   gem.add_development_dependency "rake"
   gem.add_development_dependency "standardrb"
--- a/lib/cable_ready.rb
+++ b/lib/cable_ready.rb
@@ -6,7 +6,6 @@ require "cable_ready/version"
 require "cable_ready/broadcaster"
 require "cable_ready/config"
 
-
 module CableReady
   class Engine < Rails::Engine
   end
--- a/lib/cable_ready.rb
+++ b/lib/cable_ready.rb
@@ -4,8 +4,14 @@ require "rails/engine"
 require "active_support/all"
 require "cable_ready/version"
 require "cable_ready/broadcaster"
+require "cable_ready/config"
+
 
 module CableReady
   class Engine < Rails::Engine
+  end
+
+  def self.config
+    @config ||= Config.new
   end
 end
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -43,10 +43,10 @@ module CableReady
 
       if dispatch_operations.key? :dispatch_event
         dispatch_operations[:dispatch_event].each do |event|
-          event[:detail]&.deep_transform_keys! { |key| case_for_detail(key) } #CableReady.config
-          event.transform_keys!{ |key| key.to_s.camelize(:lower) }
+          event[:detail]&.deep_transform_keys! { |key| case_for_detail(key) } # CableReady.config
+          event.transform_keys! { |key| key.to_s.camelize(:lower) }
         end
-        dispatch_operations.transform_keys!{ |key| key.to_s.camelize(:lower) }
+        dispatch_operations.transform_keys! { |key| key.to_s.camelize(:lower) }
         operations.merge! dispatch_operations
       end
     end
@@ -61,7 +61,7 @@ module CableReady
         key.to_s.dasherize
       when "camel"
         key.to_s.camelize(:lower)
-      when "keep" #keeps original
+      when "keep" # keeps original
         key.to_s
       else
         key.to_s.camelize(:lower)
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -71,6 +71,5 @@ module CableReady
     def config
       CableReady.config
     end
-
   end
 end
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -15,14 +15,14 @@ module CableReady
 
     def broadcast(clear)
       operations.select! { |_, list| list.present? }
-      operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
+      handle_operations_keys! operations
       ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
       reset if clear
     end
 
     def broadcast_to(model, clear)
       operations.select! { |_, list| list.present? }
-      operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
+      handle_operations_keys! operations
       identifier.broadcast_to model, "cableReady" => true, "operations" => operations
       reset if clear
     end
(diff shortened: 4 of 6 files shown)


leastbad:
Hey @anaice!

I finally had a chance to review. There are a few things that we're going to have to change up, because now that I see the implementation, I'm no longer convinced that a configuration is the best approach for this problem.

First thing first, let's start by temporarily forgetting my last statement and looking at how you did the config. You brought in a dependency to manage a config file, but CableReady already has support for an initializer. This is why I mentioned that it was important to consult #47. Instead of using anyway_config, we can just add additional settings to the initializer.

I also don't think we should attempt to provide pascal or kebab at this time. camelCase is clearly the default, and I agree that there's value in offering snake_case, too. So this means that however we attack this, we can reduce it to a ternary with the pseudocode:

`operations.deep_transform_keys! { |key| camel ? key.to_s.camelize(:lower) :  key.to_s.underscore }`

Here's where I'm at with this: I feel awkward about setting a precedent for individual operations, which on the server are generally treated equally. We should expect that people might want other methods to return detail objects that have snake case as well.

My sense is that if someone wants to send a broadcast with snake case, we should let them set that as a parameter on a broadcast. There is already a precedent for this; both `broadcast` and `broadcast_to` methods have an optional named parameter called `clear` that gets passed to the channel on channels.rb lines 56 and 66 as a boolean.

What I think makes more sense is to do this in channels.rb:
```rb
    def broadcast(*identifiers, clear: true, camel: true)
      @channels.values
        .reject { |channel| identifiers.any? && identifiers.exclude?(channel.identifier) }
        .select { |channel| channel.identifier.is_a?(String) }
        .tap do |channels|
          channels.each { |channel| @channels[channel.identifier].broadcast(clear: clear, camel: camel) }
          channels.each { |channel| @channels.except!(channel.identifier) if clear }
        end
    end

    def broadcast_to(model, *identifiers, clear: true, camel: true)
      @channels.values
        .reject { |channel| identifiers.any? && identifiers.exclude?(channel.identifier) }
        .reject { |channel| channel.identifier.is_a?(String) }
        .tap do |channels|
          channels.each { |channel| @channels[channel.identifier].broadcast_to(model, clear: clear, camel: camel) }
          channels.each { |channel| @channels.except!(channel.identifier) if clear }
        end
    end
```
and then in channel.rb:
```rb
    def broadcast(clear:, camel:)
      operations.select! { |_, list| list.present? }
      operations.deep_transform_keys! { |key| camel ? key.to_s.camelize(:lower) :  key.to_s.underscore }
      ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
      reset if clear
    end

    def broadcast_to(model, clear:, camel:)
      operations.select! { |_, list| list.present? }
      operations.deep_transform_keys! { |key| camel ? key.to_s.camelize(:lower) :  key.to_s.underscore }
      identifier.broadcast_to model, "cableReady" => true, "operations" => operations
      reset if clear
    end
```

This would allow you as developer to do something like:
```rb
cable_ready["stream"].dispatch_event(detail: {user_name: "leastbad"}, selector: "body", name: "my_event")
cable_ready.broadcast(camel: false)
```
and this would allow the granular ability to override any operation with snake case details on a per-broadcast basis.

What do you think, @hopsoft?

leastbad:
Late thought: another variation on this could be to make the falsy condition simply not do any modifications, so the developer could have direct control of exactly what gets sent down the wire. This might actually be an even better approach.

channel.rb:
```rb
    def broadcast(clear:, camel:)
      operations.select! { |_, list| list.present? }
      operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) } if camel
      ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
      reset if clear
    end

    def broadcast_to(model, clear:, camel:)
      operations.select! { |_, list| list.present? }
      operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) } if camel
      identifier.broadcast_to model, "cableReady" => true, "operations" => operations
      reset if clear
    end
```

anaice:
> Hey @anaice!
> 

Hey @leastbad!
First of all I would like to thank you for your time helping me with this. 🙏

> I finally had a chance to review. There are a few things that we're going to have to change up, because now that I see the implementation, I'm no longer convinced that a configuration is the best approach for this problem.
> 
In my opinion your original idea of ​​configuration is good, because generally when you want a case pattern the application will stick to it. The best option in my opinion would be the configuration and also an optional override per-broadcast operation. But I agree that we can do with the parameter only.

> First thing first, let's start by temporarily forgetting my last statement and looking at how you did the config. You brought in a dependency to manage a config file, but CableReady already has support for an initializer. This is why I mentioned that it was important to consult #47. Instead of using anyway_config, we can just add additional settings to the initializer.
> 
Yes. Sorry about that

> I also don't think we should attempt to provide pascal or kebab at this time. camelCase is clearly the default, and I agree that there's value in offering snake_case, too. So this means that however we attack this, we can reduce it to a ternary with the pseudocode:
> 
> `operations.deep_transform_keys! { |key| camel ? key.to_s.camelize(:lower) : key.to_s.underscore }`
> 
I understand. I just thought that if we were to give the config  we could have more options, giving more control to the developer. I also think that if we are going to use the ternary we should choose not to make any changes to the keys (the "keep" option of the PR). This way the event detail payload will not be changed: if the detail has `user_id` or `myUser_Id` the format will be maintained.


> Here's where I'm at with this: I feel awkward about setting a precedent for individual operations, which on the server are generally treated equally. We should expect that people might want other methods to return detail objects that have snake case as well.
> 
> My sense is that if someone wants to send a broadcast with snake case, we should let them set that as a parameter on a broadcast. There is already a precedent for this; both `broadcast` and `broadcast_to` methods have an optional named parameter called `clear` that gets passed to the channel on channels.rb lines 56 and 66 as a boolean.
> 
> What I think makes more sense is to do this in channels.rb:
> 
> ```ruby
>     def broadcast(*identifiers, clear: true, camel: true)
>       @channels.values
>         .reject { |channel| identifiers.any? && identifiers.exclude?(channel.identifier) }
>         .select { |channel| channel.identifier.is_a?(String) }
>         .tap do |channels|
>           channels.each { |channel| @channels[channel.identifier].broadcast(clear: clear, camel: camel) }
>           channels.each { |channel| @channels.except!(channel.identifier) if clear }
>         end
>     end
> 
>     def broadcast_to(model, *identifiers, clear: true, camel: true)
>       @channels.values
>         .reject { |channel| identifiers.any? && identifiers.exclude?(channel.identifier) }
>         .reject { |channel| channel.identifier.is_a?(String) }
>         .tap do |channels|
>           channels.each { |channel| @channels[channel.identifier].broadcast_to(model, clear: clear, camel: camel) }
>           channels.each { |channel| @channels.except!(channel.identifier) if clear }
>         end
>     end
> ```
> 
> and then in channel.rb:
> 
> ```ruby
>     def broadcast(clear:, camel:)
>       operations.select! { |_, list| list.present? }
>       operations.deep_transform_keys! { |key| camel ? key.to_s.camelize(:lower) :  key.to_s.underscore }
>       ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
>       reset if clear
>     end
> 
>     def broadcast_to(model, clear:, camel:)
>       operations.select! { |_, list| list.present? }
>       operations.deep_transform_keys! { |key| camel ? key.to_s.camelize(:lower) :  key.to_s.underscore }
>       identifier.broadcast_to model, "cableReady" => true, "operations" => operations
>       reset if clear
>     end
> ```
> 
> This would allow you as developer to do something like:
> 
> ```ruby
> cable_ready["stream"].dispatch_event(detail: {user_name: "leastbad"}, selector: "body", name: "my_event")
> cable_ready.broadcast(camel: false)
> ```
> 
> and this would allow the granular ability to override any operation with snake case details on a per-broadcast basis.
> 
Even though I still prefer configuration. I’m ok with a per-broadcast option. 

One important thing: we want to change only the detail payload for :dispatch_event operations, so the transformation of keys must be done differently. 
Initially I just used deep_transform_keys, but this changes the whole object graph which ended up breaking the JS part of the lib, as it uses camelCase, for example: dispatch_event does not work, dispatchEvent works. Therefore, the entire object must be kept camel, only the content of the detail must be configurable. That's why we have:

```ruby
# lib/cable_ready/channel.rb
#...
   def broadcast(clear)
      operations.select! { |_, list| list.present? }
      # operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
      handle_operations_keys! operations
      ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
      reset if clear
    end

    def handle_operations_keys!(operations)
      dispatch_operations = operations.extract! :dispatch_event

      if dispatch_operations.key? :dispatch_event
        dispatch_operations[:dispatch_event].each do |event|
          event[:detail]&.deep_transform_keys! { |key| case_for_detail(key) } #CableReady.config
          event.transform_keys!{ |key| key.to_s.camelize(:lower) }
        end
        dispatch_operations.transform_keys!{ |key| key.to_s.camelize(:lower) }
        operations.merge! dispatch_operations
      end
    end

    def case_for_detail(key)
      case config.detail_case
      when "snake"
        key.to_s.underscore
      when "pascal"
        key.to_s.camelize
      when "kebab"
        key.to_s.dasherize
      when "camel"
        key.to_s.camelize(:lower)
      when "keep" #keeps original
        key.to_s
      else
        key.to_s.camelize(:lower)
      end
    end
#...
```
I've a running application using CableReady with the PR version:
object-graph-js-evt-cable-ready

`operations.deep_transform_keys!` all way down will result in dispatch_event

> What do you think, @hopsoft?

@leastbad and @hopsoft  I will follow your discussion on this.
Thanks again for your time and for this excellent gem!

anaice:
> Late thought: another variation on this could be to make the falsy condition simply not do any modifications, so the developer could have direct control of exactly what gets sent down the wire. This might actually be an even better approach.
> 
> channel.rb:
> 
> ```ruby
>     def broadcast(clear:, camel:)
>       operations.select! { |_, list| list.present? }
>       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) } if camel
>       ActionCable.server.broadcast identifier, "cableReady" => true, "operations" => operations
>       reset if clear
>     end
> 
>     def broadcast_to(model, clear:, camel:)
>       operations.select! { |_, list| list.present? }
>       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) } if camel
>       identifier.broadcast_to model, "cableReady" => true, "operations" => operations
>       reset if clear
>     end
> ```

I agree with this. The developer is responsible for the detail payload.

leastbad:
Ahh, that's what I get for proposing code in a textarea! Good catch... so, more code would be required to modify only the details hash.

I still think that we can vastly simplify things.

camelCase is the default and will remain so. Not only are we not planning to break existing deployments and functionality, JS devs do expect to work with camelCase values. Kebab case values aren't even valid in JS and Pascal case values are, like in Ruby, used for class names. Snake case values are handy for Rails devs, so again, I have your back on being able to opt-in.

I do have to defer to @hopsoft regarding whether he'd be more preferential to a global configuration or a named parameter to broadcast.

anaice:
@leastbad 
You're right. Taking a step back I really think we can simplify things. Only a configuration and/or per-broadcast parameter will work: `camelize_detail`, a falsy condition to simply not do any modifications.
I believe that the detail payload should be chosen by the developer. But camel case is mandatory to maintain backwards compatibility, as you said it must be the default.

leastbad:
Hey @anaice! Hope that you're good.

Lots has changed since this PR was submitted, and it's safe to say that we're probably not going to be accepting this change. Still, we sincerely appreciate the contribution, regardless.

PR #63, opened 2020-07-23
Provide an async perform method

julianrubisch:
In certain situations it can be preferable to `await` until morphdom operations are completed asynchronously to free the main thread to do other work. 

One example would be to trigger events that bubble up the DOM without waiting for the morphs to finish.

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -329,4 +329,17 @@ const perform = (
   }
 }
 
-export default { perform, isTextInput, DOMOperations }
+const performAsync = (
+  operations,
+  options = { emitMissingElementWarnings: true }
+) => {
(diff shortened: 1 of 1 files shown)


hopsoft:
This looks good. Will you update the PR description to include a usage example etc... as a future reference to help with the docs?

leastbad:
This seems like a great idea, but it does mean that things suddenly happen in a less predictable order. It's nice to be able to decided to put this to work for a channel, but I would urge against making it the default in SR, for example. 😃

julianrubisch:
That was not my intention at all. I‘m thinking of futurism here...

leastbad:
I can't wait to benchmark both versions. Exciting times.

Issue #64, opened 2020-07-27
Operations go missing before broadcasting


barrywoolgar:
Hello

First off, thank you for this excellent gem. I'm running 4.3.0, and developing on OSX with Ruby 2.7.1 and Rails 6 master.

I believe I've found a concurrency problem in the `CableReady::Channels` singleton.

My app will occasionally enqueue 1000s of Sidekiq jobs in a very short period of time, each of which will generate a small number of morph operations and then broadcast them. A rather small percentage of these jobs will consistently fail inside cable_ready, be re-queued, and later succeed once the queue has shrunk.

The error is: 

    NoMethodError (undefined method `broadcast' for nil:NilClass):
    cable_ready-4.3.0/lib/cable_ready/channels.rb:56:in `block (2 levels) in broadcast'

And occurs on here:

https://github.com/hopsoft/cable_ready/blob/master/lib/cable_ready/channels.rb#L56

    channels.each { |channel| @channels[channel.identifier].broadcast(clear) }

From what I can see with some debug statements is the `@channels` instance var is an empty hash on line 56, despite having to have had keys/values on line 52 to `select` on line 54.

Unfortunately I'm not very familiar with this kind of problem, so not sure if there's anything I can do to mitigate it from my application code. I'm currently working around it by forcing Sidekiq to have a concurrency of 1, but I'm not sure how performant that would be in production.

My job class for reference:

    class BroadcastJob < ActiveJob::Base
      queue_as :default
      sidekiq_options retry: 5
      include CableReady::Broadcaster

      def perform(object)
        cable_ready["subscription-#{object.subscription_id}"].morph(
          selector: "#object_#{object.id}",
          html: renderer.render(partial: "subscriptions/row", locals: { object: object })
        )

        cable_ready["month-#{object.month_id}"].morph(
          selector: "#object_#{object.id}",
          html: renderer.render(partial: "months/row", locals: { object: object })
        )

        cable_ready.broadcast
      end

      private

      def renderer
        @renderer ||= ApplicationController.renderer.new(https: true)
      end
    end

Many thanks for any assistance you can provide!

barrywoolgar:
On further consideration, I imagine the "missing" operations may have still been broadcast, just by a different process to the one that added it to @channels. Though perhaps it's just as possible that the `except!` on line 57 could cause actual loss if another process added an operation in the meantime.

leastbad:
Hey Barry,

[...]

PR #65, opened 2020-07-28
Add a shared mutex to avoid concurrency issues in Channels

barrywoolgar:
This resolves the issue discussed in #64 where the @channels object may be modified out-of-sync in a multi-threaded environment, potentially resulting in premature broadcasting or loss of operations before broadcast.


barrywoolgar:
I believe this is failing due to a StandardRB configuration problem, but please do let me know if you need me to fix anything.

leastbad:
This is going to have to wait until @hopsoft is back from vacation. Thanks for your patience!

Issue #70, opened 2020-09-07
shouldMorph doesn't work well with some form elements


adank92:
First of all, I want to thank you for this awesome gem (and for Stimulus Reflex too). It has been fun using them!

I noticed that performing a full page morph is causing some issues with `select` and `input` tags. The first issue I ran into is that select tags that have a "blank" option selected are being cleared out even if their values are not changed. After some debugging I was able to connect these issues with the `shouldMorph` method.

I've created [this repo](https://github.com/adank92/cable_ready_form_element_issues/) to show the issues described here.

### First issue: select tags are cleared out

![select](https://user-images.githubusercontent.com/5948052/92415871-3b5f1500-f131-11ea-8bf0-e41d175578ca.gif)

**Code**
```
# app/views/tests/index.html.erb
<%= select_tag :foo, raw('<option>Test</option>'), prompt: 'Blank Option' %>
<%= select_tag :bar, raw('<option>Test</option>'), prompt: 'Blank Option', data: { reflex: 'change->test#change' } %>
```
```
# app/controllers/tests_controller.rb
class TestsController < ApplicationController
  def index
  end
end
```
```
# app/reflexes/test_reflex.rb
class TestReflex < ApplicationReflex
  def change
  end
end
```
**This is what morphdom does if we don't use `shouldMorph` at all:**

```
// This line gets called at some point, sending the current blank option and the blank option
// from the backend, which are the same
syncBooleanAttrProp(fromEl, toEl, "selected");
```
[source](https://github.com/patrick-steele-idem/morphdom/blob/master/src/specialElHandlers.js#L35)
```
// This method will set the "selected" to the blank option
function syncBooleanAttrProp(fromEl, toEl, name) {
  if (fromEl[name] !== toEl[name]) {
    fromEl[name] = toEl[name];
    if (fromEl[name]) {
      fromEl.setAttribute(name, "");
    } else {
      fromEl.removeAttribute(name);
    }
  }
}
```
[source](https://github.com/patrick-steele-idem/morphdom/blob/master/src/specialElHandlers.js#L1)
```
// Later on this method is called to morph the select input
SELECT: function (fromEl, toEl) {
    if (!toEl.hasAttribute("multiple")) {
      var selectedIndex = -1;
[...]
          if (nodeName === "OPTION") {

            // There's a child with selected here! It will change selectedIndex from -1 to 0

            if (curChild.hasAttribute("selected")) {
              selectedIndex = i;
              break;
            }
            i++;
[...]

PR #71, opened 2020-09-08
Removed isEqualNode comparison

adank92:
Proposal to fix https://github.com/hopsoft/cable_ready/issues/70

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -79,9 +79,6 @@ const getClassNames = names => Array(names).flat()
 // SEE: https://github.com/patrick-steele-idem/morphdom#morphdomfromnode-tonode-options--node
 //
 const shouldMorph = permanentAttributeName => (fromEl, toEl) => {
-  // Skip nodes that are equal:
-  // https://github.com/patrick-steele-idem/morphdom#can-i-make-morphdom-blaze-through-the-dom-tree-even-faster-yes
-  if (fromEl.isEqualNode(toEl)) return false
   if (!permanentAttributeName) return true
 
   const permanent = fromEl.closest(`[${permanentAttributeName}]`)


hopsoft:
I think we should retain the `isEqualNode` check for performance reasons; however, I'd be ok to skip this check for `SELECT` and `OPTION` elements. @adank92 Do you think this compromise would address the problems you faced in #70?

leastbad:
That is a clever suggestion, if the issue is restricted to just `select`/`option`.

adank92:
@hopsoft @leastbad that would work for the first issue, but for the second one we would also need to skip `INPUT`

leastbad:
Talk to me like I'm distracted and less confident... what is it about form elements that they need special care? Is it because the user can change their state?

Should `TEXTAREA` be added to the list?

Are there others we're not thinking about because "nobody" ever uses them?

What are the gotchas? eg. what about the `input type="numeric"` that is forced to display the value `abc`, putting it in violation of the half-baked pattern validation built into some elements? is "equal but not valid" something that matters in this context? I'm just trying to brainstorm and still feeling haunted by the abstract thinking required to get to an outcome on https://github.com/patrick-steele-idem/morphdom/pull/206

adank92:
>Talk to me like I'm distracted and less confident... what is it about form elements that they need special care? Is it because the user can change their state?

`SELECT` elements need to have one option with the `selected` attribute set in order for morphdom to work as expected. The library will ensure this by running code to sync the attribute (`syncBooleanAttrProp(fromEl, toEl, "selected");`). Using `isEqualNode` prevents this from happening. This is explained more in detail in the original issue #70

`INPUT` elements do not update their `value` attribute in the HTML definition when users type in it. If the page is loaded with an empty input, the user types in something, and then the backend returns an empty input again, `ìsEqualNode` will always return true because both HTML elements are still the same.

Screenshot_20201115_165401

Screenshot_20201115_165935

> Should TEXTAREA be added to the list?

Yes, `TEXTAREA` has the same issue as `INPUT`. If we want to retain `isEqualNode`, it should be skipped too.

> Are there others we're not thinking about because "nobody" ever uses them?

There might be others but I can only confirm these 3 elements. I don't think the other form elements in this list would have issues https://www.w3schools.com/html/html_form_elements.asp. But then again, there's a reason why morphdom doesn't use `isEqualNode`.

> What are the gotchas? eg. what about the input type="numeric" that is forced to display the value abc, putting it in violation of the half-baked pattern validation built into some elements? is "equal but not valid" something that matters in this context? I'm just trying to brainstorm and still feeling haunted by the abstract thinking required to get to an outcome on patrick-steele-idem/morphdom#206

Numeric inputs suffer from the same issue, their values are not updated if changed in the backend. I've tested it without `isEqualNode` and it works fine. If the backend clears the value, the UI clears it. If the backend changes the number, the UI changes it. And if the backend changes the value to a string, the UI just clears the value out.

leastbad:
Interestingly, in my testing with a draft PR that doesn't use `isEqualNode` if the tagName is an input/select/textarea...

I am finding that it's not `select` that is the problem, but `option`. So far as I can tell, @adank92's first MCVE works fine if I filter for `option` but still breaks if I filter for `select`.

Does that make sense to you, Adan?

Works:
```js
  if (fromEl.tagName != 'OPTION' && fromEl.isEqualNode(toEl)) return false
```

Doesn't work:
```js
  if (fromEl.tagName != 'SELECT' && fromEl.isEqualNode(toEl)) return false
```

I was/am worried that this bug will impact `DATALIST` elements as well, but if indeed it's the `OPTION` that is relevant, this could kill two birds with one stone.

adank92:
Yes, you are right. `select` itself is not a problem, it's `option`. Since `morphdom` can't find an `option` with a `selected` attribute, it ends up resetting the select.

Thanks for taking care of it in a new PR.

PR #73, opened 2020-09-25
Add support for pushState

hopsoft:


leastbad:
This seems to break the docs...

I would just use GitBook after the fact?

leastbad:
I think this operation is going to be a game-changer.

PR #78, opened 2020-11-05
method chaining

leastbad:
@coderberry requested that CableReady operations be chainable. That was actually so easy, I'm a bit sad we didn't do this all along. I just had to modify `Channel#add_operation` to return `self`.

```ruby
cable_ready["MyChannel"].morph(selector: "body", content: "blah").dispatch_event(name: "boom", detail: {})
```

The logical progression from there was to be able to conclude a chain with `broadcast` or `broadcast_to`. This was a little more complicated, as the `Channel` class already had `broadcast` and `broadcast_to` methods. Arguably, this was unnecessarily confusing and disambiguating to `channel_broadcast` and `channel_broadcast_to` helps make the code easier to follow.

I changed the similarly confusing `Channel#add_operation` to `Channel#enqueue_operation` for similar reasons.

Anyhow, with methods renamed, we now have new `Channel#broadcast` and `Channel#broadcast_to` methods which implicitly pass the current channel identifier and/or model to update so there's no ambiguity about what is being broadcast. `Channel#broadcast` accepts an optional boolean for the `clear` argument, defaulting to `true`. `Channel#broadcast_to` requires a model instance argument and allows the optional boolean for the `clear` argument, again defaulting to `true`.

```ruby
cable_ready["MyChannel"].morph(selector: "body", content: "blah").dispatch_event(name: "boom", detail: {}).broadcast_to(User.find(params[:id]))
```

### DIFF
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -13,24 +13,33 @@ module CableReady
       end
     end
 
-    def broadcast(clear)
+    def channel_broadcast(clear)
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
       ActionCable.server.broadcast identifier, {"cableReady" => true, "operations" => operations}
       reset if clear
     end
 
-    def broadcast_to(model, clear)
+    def channel_broadcast_to(model, clear)
       operations.select! { |_, list| list.present? }
       operations.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
       identifier.broadcast_to model, {"cableReady" => true, "operations" => operations}
       reset if clear
     end
 
+    def broadcast(clear = true)
+      cable_ready.broadcast(identifier, clear)
+    end
+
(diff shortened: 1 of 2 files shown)


hopsoft:
It wasn't obvious to me where the reference to `cable_ready` in `CableReady::Channel` was defined ( `lib/cable_ready/channel.rb+30-37`), so I setup a quick test to force that code path.

This is what I tested with from inside of a reflex.

```ruby
cable_ready[stream_name]
  .push_state(url: new_uri.to_s)
  .broadcast
```

This produced the following error.

```
StimulusReflex::Channel Failed to invoke Search#search! http://localhost:3000/feeds?query=m undefined local variable or method `cable_ready' for #<CableReady::Channel:0x00007f93e3c82070>
```

This may work for some code paths, but it failed for the scenario above.

hopsoft:
Note, I just verified that this commit has resolved the bug. https://github.com/hopsoft/cable_ready/commit/ab168635eb619aed5a824099987aa5deccca658d

PR #79, opened 2020-11-13
Move package.json to root and prepare for v4.4.0.pre3

hopsoft:


bin/standardize
```diff
@@ -2,4 +2,4 @@
 
 bundle exec magic_frozen_string_literal
 bundle exec standardrb --fix
-cd ./javascript && yarn run prettier-standard cable_ready.js
+yarn run prettier-standard ./javascript/*.js
```
ParamagicDev, line 5: In case you get any nested files in the future, you may want this to be:

./javascript/**/*.js

ParamagicDev (changes requested):
Looks good from here. May just want to fix glob pattern for the future.

Also make sure to update the version in the package.json

PR #80, opened 2020-11-16
only compare isEqualNode for non-inputs

leastbad:
This partially undoes #71 by installing an additional guard. We don't want to trust `isEqualNode` comparisons for input/textarea/option elements for the reasons specified in #70.


leastbad:
@adank92 here is my proposed fix for #70

hopsoft (approved):

PR #83, opened 2020-12-01
Add support for sessionStorage and localStorage

hopsoft:
# Add support for sessionStorage and localStorage

## Example Usage

```ruby
cable_ready[stream_name]
  .storage_set_item(key: "local-example", value: "a value in local storage")
  .storage_set_item(key: "session-example", value: "a value in session storage", type: "session")
  .broadcast

cable_ready[stream_name]
  .storage_remove_item(key: "local-example")
  .storage_remove_item(key: "session-example", type: "session")
  .broadcast
```

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -119,6 +119,24 @@ const DOMOperations = {
     dispatch(document, 'cable-ready:after-push-state', config)
   },
 
+  // Storage .................................................................................................
+
+  storageSetItem: config => {
+    const { key, value, storageType } = config
+    dispatch(document, 'cable-ready:before-storage-set-item', config)
+    storage = storageType === 'session' ? sessionStorage : localStorage
+    storage.setItem(key, value)
+    dispatch(document, 'cable-ready:after-storage-set-item', config)
+  },
+
+  storageRemoveItem: config => {
+    const { key, storageType } = config
+    dispatch(document, 'cable-ready:before-storage-remove-item', config)
+    storage = storageType === 'session' ? sessionStorage : localStorage
+    storage.removeItem(key)
+    dispatch(document, 'cable-ready:after-storage-remove-item', config)
+  },
+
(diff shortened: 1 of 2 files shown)


leastbad:
Woah, that's interesting.

I don't have experience with either session or local storage; it seems like localStorage is the preferred option. I'll have to do some MDN digging to familiarize myself so I can document this.

javascript/cable_ready.js
```diff
@@ -119,6 +119,24 @@ const DOMOperations = {
     dispatch(document, 'cable-ready:after-push-state', config)
   },
 
+  // Storage .................................................................................................
+
+  storageSetItem: config => {
+    const { key, value, storageType } = config
+    dispatch(document, 'cable-ready:before-storage-set-item', config)
+    storage = type === 'session' ? sessionStorage : localStorage
```
marcoroth, line 127: Shouldn't this be `storageType`?

```suggestion
    storage = storageType === 'session' ? sessionStorage : localStorage
```

Or otherwhise `type` on line L125

javascript/cable_ready.js
```diff
@@ -119,6 +119,24 @@ const DOMOperations = {
...
+    storage.setItem(key, value)
+    dispatch(document, 'cable-ready:after-storage-set-item', config)
+  },
+
+  storageRemoveItem: config => {
+    const { key, storageType } = config
+    dispatch(document, 'cable-ready:before-storage-remove-item', config)
+    storage = type === 'session' ? sessionStorage : localStorage
```
marcoroth, line 135: Same here

hopsoft: Ah good catch. I renamed to `type` but didn't update everywhere. Will fix.

leastbad:
Is there anything left to be said, here? This seems exciting in prospect, regardless of whether I immediately understand how I might use it or not. 😀

PR #84, opened 2020-12-12
Add more detail to the CableReady error message

marcoroth:
# Type of PR
Enhancement

## Description

Currently it just states `CableReady detected an error in morph! Object doesn't support this action` which makes it pretty hard to debug.

#### Before

Screenshot 2020-12-12 at 17 56 11

#### After
Now it's shown as an error (which it obviously is), shows the source where the error was thrown and prints the error separately so you can have a look the the error/stack-trace. 

Screenshot 2020-12-12 at 19 31 36

## Why should this be added

Makes it easier to debug and DX.

## Checklist

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -365,7 +365,7 @@ const perform = (
         } catch (e) {
           if (detail.element) {
             console.error(
-              `CableReady detected an error in ${name}! ${e.message}. If you need to support older browsers make sure you've included the corresponding polyfills. Check this out for more details: https://docs.stimulusreflex.com/setup#polyfills-for-ie11.`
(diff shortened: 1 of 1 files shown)


javascript/cable_ready.js
```diff
@@ -363,12 +363,16 @@ const perform = (
           if (detail.element || options.emitMissingElementWarnings)
             DOMOperations[name](detail)
         } catch (e) {
-          if (detail.element)
-            console.log(`CableReady detected an error in ${name}! ${e.message}`)
-          else
+          if (detail.element) {
+            console.error(
+              `CableReady detected an error in ${name}! ${e.message}. If you need to support older browsers make sure you've included the corresponding polyfills. Check this out for more details: https://docs.stimulusreflex.com/setup#polyfills-for-ie11.`
```
hopsoft, line 368: I think we can omit this part of the message. `Check this out for more details:`

hopsoft (approved):

PR #86, opened 2020-12-13
changed storage method names, added clear flag

leastbad:
`set_storage` and `remove_storage` follow the naming convention used by every other operation.

I added a new optional `clear` flag to the `remove_storage` method that optionally clears all keys.

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -121,20 +121,20 @@ const DOMOperations = {
 
   // Storage .................................................................................................
 
-  storageSetItem: config => {
+  setStorage: config => {
     const { key, value, type } = config
-    dispatch(document, 'cable-ready:before-storage-set-item', config)
+    dispatch(document, 'cable-ready:before-set-storage', config)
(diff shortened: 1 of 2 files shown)


marcoroth:
Wouldn't it make more sense to provide an extra `clear_storage` method that just clears the storage since it's also a method on it's own in the browser API?

leastbad:
Maybe! I would defer to Nate's preference here - clearing is definitely a kind of removal.

Fwiw I'm definitely influenced by the reality that this seems like it'd be usually very rarely.

hopsoft:
I like the semantic change here for consistency, but don't think we should drop the `item` suffix. I also like the idea of adding `clearStorage`. One other nitpick that I'd like to address while we're at it is to move

```js
const storage = type === 'session' ? sessionStorage : localStorage
```

directly under the line where we destructure vars out of the config.

leastbad:
Sounds good to me! I updated `channels.rb` with the new method names.

hopsoft (approved):

PR #87, opened 2020-12-16
pluggable before/after morph callbacks

leastbad:
Continuing with the developer happiness theme, I've noticed that we're currently failing two demographics: people who want to use AlpineJS and people who want to use Shoelace. Both tools require things to happen before and after morphing which require substantial changes to the library to achieve. This PR is those changes; it solves current and future asks by providing a modular system for pluggable `onBeforeElUpdated` and `onElUpdated` callbacks that function similarly to Rack middleware in concept.

`CableReady` now exports `shouldMorphCallbacks` and `didMorphCallbacks`. Both are **arrays of functions** that can be `push`ed on to, or in the case of `shouldMorphCallbacks`, `slice`d if necessary.

`shouldMorph` and its new sister callback `didMorph` both execute each function in their respective arrays sequentially, from `0` to `n`.

`didMorphCallbacks` starts as `[]` while `shouldMorphCallbacks` starts as `[verifyNotMutable, verifyNotPermanent]`.

`verifyNotMutable` is the mutable tags/`isEqualNode` handling that used to be at the top of `shouldMorph`, modified to return `true` if everything is in order.

`verifyNotPermanent` is the permanent attribute handling that was previously in `shouldMorph`. It could theoretically be `slice`d from `beforeConditions` for performance reasons, if you didn't ever use `data-reflex-permanent`.

The `morph` method itself has been modified to pass the full `details` object to the callbacks. It always felt a bit weird to have the `shouldMorph` callback know what `permanentAttributeName` is.

I have tidied up the `cable_ready.js` monolith by creating `enums.js`, `callbacks.js` and `utils.js` which should be self-explanatory. I did have to export `activeElement` for this to work, but this shouldn't impact anything.

Let's enable Alpine.js:
```js
// application.js

const enableAlpine(detail, fromEl, toEl) {
  if (fromEl.__x) { window.Alpine.clone(fromEl.__x, toEl) }
  return true
}

CableReady.shouldMorphCallbacks.push(enableAlpine)
```

Let's fix Shoelace:
```js
// application.js

const fixShoelace(detail, el) {
  if (el.tagName.startsWith('SL-')) el.classList.add('hydrated')
}

CableReady.didMorphCallbacks.push(fixShoelace)
```


leastbad:
If @julianrubisch prefers, we could provide methods to add/remove functions to/from the `beforeConditions` and `afterEffects` arrays, and make the arrays private. I guess a hypothetical `remove()` would just get rid of the last element of the array - in this case, `verifyNotPermanent`.

julianrubisch:
The remark about removing mandatory checks is indeed a good indicator that we should provide custom accessors. 

Also, I was a bit puzzled by the naming of _condition_. Technically of course correct, but it sent me down the wrong mental path initially. Maybe _check_ might indeed be clearer?

leastbad:
> The remark about removing mandatory checks is indeed a good indicator that we should provide custom accessors.

Could you be more specific? What remark are you referring to?

I'm open to changing `conditions` -> `checks`; let's wait for everyone to have a look so I'm not chasing my tail (only want to change it once). 😀

julianrubisch:
> I guess a hypothetical remove() would just get rid of the last element of the array - in this case, verifyNotPermanent.

👆🏻

leastbad:
Gotcha. Yeah, I was just thinking out potential API surface ideas.

Keep in mind, people who are adding operations are also already manipulating the DOMOperations object; if we were going to be consistent, we'd bump the library to v5 and provide accessor methods for it as well.

hopsoft:
I like where this is going.

I propose that we rename the following items given that they're scoped to the `morph` operation.

- `beforeConditions` -> `shouldMorphCallbacks`
- `afterEffects` -> `didMorphCallbacks`

leastbad:
Thanks for the notes, Nate. I'll change those function names now.

I guess the big question I need you to weigh in on is whether we:

1. Follow the precedent set by DOMOperations and export the two arrays for people to modify. This is the current scenario.
2. Bump to v5 and replace the DOMOperations export with an accessor method, while also adding accessors for the two new arrays.
3. Leave the DOMOperations as an array people can access but create accessors for the two new arrays, which feels inconsistent.

Truthfully, I don't feel passionately about any of these options, I just want to do the work once and get it shipped!

hopsoft:
I think option 1 is perfectly fine unless we have users getting tripped up by it.

leastbad:
I think it's ready to go, then!

hopsoft (approved):

PR #88, opened 2020-12-21
rework custom operations

leastbad:
I suspect that this effectively undocumented feature (made real in #47) hasn't been working as intended for some time. I believe you can currently add operations with default behaviour. Unfortunately, custom behaviour just falls down in today's CableReady, especially where method chaining is concerned.

On the eve of real documentation emerging, I suspect people will want to make use of this feature, and this PR fixes it.

```rb
CableReady::Channels.configure do |config|
  config.add_operation :jazz_hands do |options|
    puts options.inspect
    puts "Jazz hands!"
    puts rand 1..1000
  end
end
```

Now, I'm immediately putting this into draft because in truth... this might be a solution looking for a problem.

If this PR were accepted, the behaviour would be that the code in the block would execute when the operation was added to the queue. In my original (incorrect) conception, the code would fire at the time of the broadcast. This is incorrect.

So, the question becomes: which of these behaviours make the most sense?

1. I have a use case in mind where the current behaviour (block yield at the time I add an operation to the queue) makes sense.
2. I have a use case in mind, but I would prefer the block yield happen at the time of broadcast.
3. Let's pull out this block yield stuff and just add the custom operation to the list of available operations.

Basically, I think the idea of defining functionality for server-side behaviour as part of operations is cool, but I am hard pressed to come up with an example of how this could be used.

### DIFF
--- a/lib/cable_ready/channels.rb
+++ b/lib/cable_ready/channels.rb
@@ -47,9 +47,9 @@ module CableReady
 
     def add_operation(operation)
       @operations[operation] = ->(options = {}) do
-          yield(options) if block_given?
-          enqueue_operation(operation, options)
-        end
+        yield(options) if block_given?
+        enqueue_operation(operation, options)
+      end
     end
 
     def [](identifier)
--- a/lib/cable_ready/channels.rb
+++ b/lib/cable_ready/channels.rb
@@ -45,8 +45,11 @@ module CableReady
       end
     end
 
-    def add_operation(operation, &implementation)
-      @operations[operation] = implementation || ->(options = {}) { enqueue_operation(operation, options) }
+    def add_operation(operation)
+      @operations[operation] = ->(options = {}) do
+          yield(options) if block_given?
+          enqueue_operation(operation, options)
+        end
     end
 
     def [](identifier)


hopsoft:
I'd expect the options block to run at the time I call `add_operation` in the example above. That would be more idiomatic IMO., but let's implement some real use cases to tease things out a bit more. I'd also love to hear from anyone that's using this feature.

leastbad:
As of right this moment, @n-rodriguez is the only person I know who is supplying their own custom operations. @foca might be, as well - I'm not sure.

I believe that this is because up until now it hasn't been documented. I suspect this will change quickly.

@hopsoft great news; if we accept this PR as written, that functionality is exactly what you'll get. This version of Potential Future Reality means that it'd function as a sort of callback for enqueueing an operation to be delivered.

(I am assuming that you mean the above, and not a callback for the one-time act of adding the custom operation to the available operations list, which happens in the initializer! Initializer callbacks aren't a Thing. 😀 )

So here's a maybe clever possible application: since the callback has access to the options passed to the method, and those same options show up in the `detail` object on the event when it arrives, perhaps you could use the callback to create instance-specific data or even fire off ActiveJobs that warm caches or something in advance of the operation being broadcast. 🤷

n-rodriguez:
> As of right this moment, @n-rodriguez is the only person I know who is supplying their own custom operations. @foca might be, as well - I'm not sure.

right

leastbad:
> right

Happy to say @n-rodriguez is no longer a rare duck. Several folks have used the new docs to implement their own custom operations.

My current take on this PR is that we should merge it as is, but leave this "server side behavior" "functionality" undocumented until such time as any of us can propose a realistic use case. The PR does fix code that is currently broken, replacing it with something that might not be super useful, but at least it won't potentially confuse people and lead to preventable errors.

PR #92, opened 2021-01-01
multiple selector element operations

leastbad:
**With genuine apologies for borking** #90.

This PR proposes the following changes:

1. All `selector` parameters now accept CSS queries that can return multiple elements. ~~Internally, `element` is now handled as an array. This includes the default, `[document]` and values returned via XPath.~~ `event.detail.element` is not an array unless `selectAll` is passed. Since nobody is currently using a feature that hasn't been released yet, this doesn't break SemVer.
2. There is a new `processList` utility function that handles calling operations either for the first element or all elements in the array, depending on whether the `selectAll: true` parameter has been passed. It defaults to just the first element, which should be the current behaviour.
3. I moved `activeElement = document.activeElement` to the `perform` function #DRY
4. I removed the `assignFocus`/`focusSelector` option from `set_value` since it's not supposed to be there and wasn't documented.
5. I moved the object destructuring to below the `before-operation` callback, giving the developers the opportunity to implement event handlers which change the parameters of the operation. For example, we had a user request to be able to change the `html` in a `before-inner-html` callback today. This is a power-user move, but it's a power-user library.
6. I added the ability to cancel operations during a `before-operation` callback by adding a `cancel: true` pair to the operation. This applies to all DOM mutation and element mutation operations **that are not morph**, which is just too complex to support cancel.
7. I consolidated all of the method parameters to be `operation` instead of `detail` or `config`. This should not impact the `detail` object produced by the `dispatch_event` operation.
8. Finally, I reordered all of the operations so that they are finally in a 1:1 order with how they are introduced in the docs, eg. by section and then alphabetical.

If anything in here doesn't pass muster, I'm happy to pull it out.

Now, I know that it's uncharacteristic of me but I would really like to implement some kind of test runner for the CableReady client. I want to be able to make these sorts of refactors without sweating that I'm going to break something dumb. I have tried a bunch of basic scenarios but it would be really great to have a long boring list of assertions to run after something like this.

~~Aside from the removal of the `assignFocus` call from `set_value` the only change that should bubble to the surface is I think harmless and minor; now that `element` is stored as an array, operations with selectors that match multiple elements will show up as an array of elements in the `detail` object of the event, even if `selectAll` is not set. I think that anyone looking for the current element is going to pull it from *event*.target, not *event*.detail.element. Right?~~

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -41,8 +41,10 @@ const DOMOperations = {
     processElements(operation, element => {
       dispatch(element, 'cable-ready:before-inner-html', operation)
       const { html, focusSelector } = operation
-      if (!operation.cancel) element.innerHTML = html
-      assignFocus(focusSelector)
+      if (!operation.cancel) {
+        element.innerHTML = html
+        assignFocus(focusSelector)
+      }
       dispatch(element, 'cable-ready:after-inner-html', operation)
     })
   },
@@ -51,9 +53,10 @@ const DOMOperations = {
     processElements(operation, element => {
       dispatch(element, 'cable-ready:before-insert-adjacent-html', operation)
       const { html, position, focusSelector } = operation
-      if (!operation.cancel)
+      if (!operation.cancel) {
         element.insertAdjacentHTML(position || 'beforeend', html)
-      assignFocus(focusSelector)
+        assignFocus(focusSelector)
+      }
       dispatch(element, 'cable-ready:after-insert-adjacent-html', operation)
     })
   },
@@ -62,9 +65,10 @@ const DOMOperations = {
     processElements(operation, element => {
       dispatch(element, 'cable-ready:before-insert-adjacent-text', operation)
       const { text, position, focusSelector } = operation
-      if (!operation.cancel)
+      if (!operation.cancel) {
         element.insertAdjacentText(position || 'beforeend', text)
-      assignFocus(focusSelector)
+        assignFocus(focusSelector)
+      }
       dispatch(element, 'cable-ready:after-insert-adjacent-text', operation)
     })
   },
@@ -80,7 +84,7 @@ const DOMOperations = {
       })
       const parent = element.parentElement
       const ordinal = Array.from(parent.children).indexOf(element)
-      if (!operation.cancel)
+      if (!operation.cancel) {
         morphdom(
           element,
           childrenOnly ? template.content : template.innerHTML,
@@ -90,7 +94,8 @@ const DOMOperations = {
             onElUpdated: didMorph(operation)
           }
         )
-      assignFocus(focusSelector)
+        assignFocus(focusSelector)
+      }
       dispatch(parent.children[ordinal], 'cable-ready:after-morph', {
         ...operation,
         content: template.content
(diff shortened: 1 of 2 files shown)


hopsoft:
This will be a great enhancement to the lib and includes some nice refactorings to boot.

> I removed the assignFocus/focusSelector option from set_value

I actually think we should keep `assignFocus` on `set_value` and just update the implementation to match the other places that are doing it.

> anyone looking for the current element is going to pull it from event.target, not event.detail.element. Right?

I'm certain that we have people destructuring element from `event.detail`

leastbad:
I just reworked how multiple elements are handled. This should resolve all concerns about breaking API compatibility.

Any thoughts on `cancel: true` or people using `before-operation` callbacks to modify payloads?

Most importantly, need @hopsoft to chime in on how/what to do about potentially removing the `after-remove` event, since the likely result is that there's no element to dispatch from. @marcoroth and I were discussing in the channel last night.

WRT `assignFocus` on `set_value`:

1. I'm pretty happy with how I've split up the operations in terms of things that impact the DOM vs things that modify one element. It sort of makes sense that if you're updating potentially many elements, that you might also want to establish where the focus should go. I'm less sure that this is true for element-specific mutations.
2. `set_value` is the only one in this category that has this capacity (and it's never been documented). Are you proposing that one operation in the category be inexplicably different, or are you suggesting that we add `assignFocus` to the other operations in the category?
3. I'd consider that've since added a standalone `set_focus` operation, and that it might make much more sense given the above to just encourage people to use it when they need it?

hopsoft:
I think we should make the `before` events cancellable and extend power to the developer. The only folks that will care will be those wanting to apply some power moves anyway. Let's let them do whatever they'd like in the before events (modify payloads, cancel, etc...)

---

I think we should dispatch `after-remove` on the document. If we have a reference to the removed element, then we should hang it off of `event.detail`.

---

OK. I'm now following your line of thinking with properly distinguishing between `DOM mutations` and `Element mutations`. It makes sense to me, but I don't think the semantics are quite right yet. Having said that, I good with leaving `assignFocus` off of `Element mutations`

Perhaps we should change the semantics to something like:

- Element tree mutations
- Element property mutations

Geesh, those are terrible... but, naming is hard. 🤷🏻‍♂️ 

---

leastbad:
Okay, I've added the ability to cancel every operation that has a `before` event, including `morph`. The only exceptions are `dispatch_event` and `console_log`, I believe.

`after-remove` is now emitted from `document`. The removed `element` reference should still be available as part of the detail object.

I changed the name of "Element Mutations" to "Element Property Mutations".

I will update the docs after merging.

Now, that just leaves testing. I have a concept in mind - it's a purpose-built app that just tests everything in the library. I don't know a better way to do it, given the unique nature and mechanism of our library.

leastbad:
Okay, I finally finished my CableReady "test suite". You can clone and run it against any local CableReady branch.

https://github.com/leastbad/stimulus_reflex_harness/tree/cr_tests

The only configuration you'll have to do is to modify the `Gemfile` to set the `branch:` on line 14 to **multi-selector** (or whatever you are testing).

Having now gone to extraordinary lengths to verify this PR, I would very much like to merge it.

leastbad:
[Harness.pdf](https://github.com/hopsoft/cable_ready/files/5826053/Harness.pdf)

PR #93, opened 2021-01-03
Remove broadcast from dispatch_event

seans84:
`broadcast` shouldn't be called here.


leastbad:
Hey, feel free to jump on #docs on the [Discord](https://discord.gg/XveN625) and explain what you mean.

Admittedly, my first reaction is "uh, yeah you do" but I'm happy to hear why I'm mistaken. 💙

seans84:
I was mistaken and using an older version of CR. This is not an issue for newer versions.

PR #107, opened 2021-02-06
chainable selectors

leastbad:
This PR proposes 4 related enhancements that brings some recent changes in StimulusReflex full circle.

1. The `dom_id` function has been moved to `CableReady::Identifiable`, which we could include in StimulusReflex's `reflex.rb` class as a way to consolidate one implementation of the function. Also, users can include this module in their own projects, although the `hash` semantics might get kind of involved (use in a client helper would require passing `hash: ""` as a parameter).
2. CableReady `selector`s can now accept `ActiveRecord::Base` and `ActiveRecord::Relation` objects, the same way the new StimulusReflex `morph` method accepts them eg. hopsoft/stimulus_reflex#436
3. When adding an operation, the method now accepts an optional `selector` as a first parameter. This is a shortcut since so many operations accept a `selector`, and saves the developer from having to type `selector: "#users"`. If a selector is provided in the `options`, it overrides a specified selector.
4. Each CableReady channel now remembers the `selector` from the previous operation, if any. This means that you can specify a selector at the beginning of a chain, and it will automatically be picked up by succeeding operations until a new selector is used, at which point that becomes the "new previous" operation for all following operations; if a new selector is used, all previously used selectors are unmodifed.

```rb
include CableReady::Indentifiable

puts dom_id(User.all, :happy) # "#happy_users"
puts dom_id(User.first) # "#user_1"

cable_ready.inner_html(selector: User.first, html: "<span>Your mother</span>")

cable_ready.inner_html("#smelly", html: "<b>All hype</b>")
cable_ready.inner_html(User.all, html: "<i>mmm</i>")

cable_ready.set_focus("#smelly").inner_html(html: "<span>I rock</span>").set_style(name: "color", value: "red").text_content(selector: User.all, text: "Bloom")
```


leastbad:
I'm still fuzzy on requires. Wouldn't it make sense to have it in the top-level file and remove it from broadcaster?

marcoroth:
I'd vote to have the requires top-level in `lib/cable_ready.rb`.

Maybe it also makes sense to switch to Zeitwerk in the future.

leastbad:
I have moved all of the `requires` in the project to `lib/cable_ready.rb` and it seems to work well in my tests.

Generally speaking, I don't really understand what would motivate `require_relative` or requiring outside of the top-level file. If you know that you're going to need it, why not keep it all in one place? This is not a rhetorical question! If there's a nuance I don't understand, I'm eager to learn.

I don't want to merge until someone either concurs or gently informs me that I was a little too aggressive.

PR #112, opened 2021-02-15
sound opt-in

### LINKED ISSUE #111: play_sound operation hijacks sound controls
In #98 a new `play_sound` operation is added. The PR describes:

> First, it uses the worlds shortest viable MP3 (93 bytes!) to overcome Apple's annoying "click to play" constraint on Mobile Safari. This process should be transparent and invisible. It does this by capturing the first interaction with the page and using that to immediately start and then stop playback; this "unlocks" the DOM object used to play the sound, document.audio.

This method has some serious drawbacks:

- With a strict Content Security Policy, the loaded audio file will generate an error on all pages:

```
Refused to load data:audio/mpeg;base64,... because it does not appear in the media-src directive of the Content Security Policy.
```

- Because the tab now has audio, all audio controls on a Mac are now linked to the tab. Clicking "play" on a keyboard starts the MP3 file instead of controlling other audio sources.
- When the page is opened on an iPhone, all other audio is paused to play the empty MP3 file. In our case it's a webview inside our native iPhone app.

We are forced to downgrade to 4.4.6, although we have no intention to use the `play_sound` operation. Maybe the audio for Mobile Safari hack should only be activated when using the `play_sound` operation?

leastbad:
It seems as though I was over-zealous in my enthusiasm for the `play_sound` operation; the bootstrap code interferes with strict CSPs and hijacks the audio context on a Mac.

This PR introduces a requirement that developers who wish to use the `play_sound` operation must place a `data-unlock-audio` attribute on their `body` tag. Otherwise, the operation will be a NOOP.

Fixes #111

### DIFF
--- a/javascript/cable_ready.js
+++ b/javascript/cable_ready.js
@@ -60,7 +60,7 @@ const performAsync = (
 }
 
 document.addEventListener('DOMContentLoaded', function () {
-  if (!document.audio) {
+  if (!document.audio && document.body.hasAttribute('data-unlock-audio')) {
     document.audio = new Audio(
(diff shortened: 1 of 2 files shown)


hopsoft (changes requested):
Looks like cleanup for `ended` is incorrect. Shouldn't it be?

```js
document.audio.removeEventListener('ended', ended)
```

leastbad:
Nice catch.

hopsoft (approved):

PR #116, opened 2021-03-07
Setup better finalizer

### LINKED ISSUE #113: Warning in Ruby 3.0
After upgrading to Ruby 3.0, the following warning gets printed when running system specs right before the first reflex invocation in a spec example:

```
.../ruby/gems/3.0.0/gems/cable_ready-4.5.0/lib/cable_ready/channel.rb:14: 
warning: finalizer references object to be finalized
```

Link to the [line in question](https://github.com/hopsoft/cable_ready/blob/6998860ca753f105c501ee564d8992a99a8e733f/lib/cable_ready/channel.rb#L14).
```
ObjectSpace.define_finalizer self, -> { CableReady.config.delete_observer config_observer }
```

hopsoft:
SEE: https://www.mikeperham.com/2010/02/24/the-trouble-with-ruby-finalizers/

Resolves #113

### DIFF
--- a/lib/cable_ready/channel.rb
+++ b/lib/cable_ready/channel.rb
@@ -4,14 +4,19 @@ module CableReady
   class Channel
     attr_reader :identifier, :enqueued_operations
 
+    def self.finalizer_for(identifier)
+      proc {
+        channel = CableReady.config.observers.find { |o| o.try(:identifier) == identifier }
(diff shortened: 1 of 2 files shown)


leastbad:
I don't know if you consider this "done", Nate, but if so, it seems like it does the trick!

PR #119, opened 2021-04-29
broadcast_later

julianrubisch:
This PR enables to delegate broadcasts to a generic `CableReadyBroadccastJob` which is autoloaded from `app/jobs`.

Using `apply!` on the `operations_payload` this was a no-brainer, the only gotcha being that class stream identifiers of course cannot be serialized as such, so I had to do `identifier.name` on the sending and `identifier.safe_constantize` on the receiving side in case a `model` is present.


leastbad:
This looks 100% ace to me.

hopsoft (approved):
Very nice.

PR #125, opened 2021-05-15
console_table operation

leastbad:
Implements https://developer.mozilla.org/en-US/docs/Web/API/Console/table as an operation.

Options:
- data
- columns

Anything passed to data is sent into the table. It expects an array or object, but fails very gracefully.
Columns defaults to `[]` and must be an array of strings if passed.

```rb
cable_ready.console_table(data: [{marco: 5, nate: 2}, {marco:4, nate: 6}], columns: ['marco']).broadcast
```

chrome_1JoQAFw3hE

Thanks to @MatheusRich for the idea!

### DIFF
--- a/javascript/operations.js
+++ b/javascript/operations.js
@@ -391,6 +391,13 @@ export default {
     })
   },
 
+  consoleTable: (operation, callee) => {
+    operate(operation, () => {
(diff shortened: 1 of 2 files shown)


hopsoft (approved):
Love it

PR #128, opened 2021-05-23
Improve install experience

leastbad:
See https://github.com/stimulusreflex/stimulus_reflex/pull/510

### DIFF
--- a/lib/cable_ready/sanity_checker.rb
+++ b/lib/cable_ready/sanity_checker.rb
@@ -31,15 +31,20 @@ class CableReady::SanityChecker
   def check_package_versions_match
     if npm_version.nil?
       warn_and_exit <<~WARN
-        Can't locate the cable_ready npm package.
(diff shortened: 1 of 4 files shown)


lib/cable_ready/sanity_checker.rb
```diff
@@ -9,40 +9,45 @@ class << self
...
 
     private
 
     def called_by_generate_config?
-      ARGV.include? "cable_ready:initializer"
+      ARGV.include?("cable_ready:initializer")
+    end
+
+    def called_by_rake?
+      caller.find { |c| c.include?("/gems/rake-") }
```
ParamagicDev, line 26: https://github.com/stimulusreflex/stimulus_reflex/pull/510#discussion_r671870456

Same comment as in the SR PR.

ParamagicDev (commented):
Same comment about Rake, otherwise, looks solid from here.

PR #137, opened 2021-06-13
Add scaffold generator

existentialmutt:
This PR adds a `rails generate cable_ready:scaffold` generator to quickly bootstrap an in-page editing interfaces for resources powered by CableReady::CableCar.

Call it just like you would rails scaffold, with a resource name and a list of name:type attributes.  It will generate the model, migration, controllers and views and ensure necessary javascript dependencies are installed (currently RailsUJS, Stimulus, and a custom Stimulus controller).

An example of the code it generates can be found here: https://github.com/existentialmutt/cable_car_messages/tree/scaffold_generator

### DIFF
--- a/lib/generators/cable_ready/scaffold/USAGE
+++ b/lib/generators/cable_ready/scaffold/USAGE
@@ -1,0 +1,45 @@
+Description:
+    Creates a CableReady-driven resource scaffold with in-page editing
+    via UJS remote links/forms and `render operations: cable_car`.
+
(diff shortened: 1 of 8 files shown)


leastbad:
I'm so excited to try this.

existentialmutt:
Excited to hear your feedback!  I just pushed an update for generating system tests.

leastbad:
Rafe and I talked; we're going to let this simmer while the we see what people do with mrujs, cable_car and SR v4.

PR #139, opened 2021-07-03
`cable-ready` element

jaredcwhite:
(**note:** updated with latest PR code…er, again…)

I've had an idea in the back of my head for a while now to enable Turbo Stream tag-like functionality but for Cable Ready operations. By updating CableCar to output HTML in addition to JSON, and then adding a custom element to CR's JavaScript which translates the HTML attributes to a JSON payload for CR to perform, by golly it actually works!

What's particularly interesting about this implementation is you don't even need to import CableReady per se. All you need to do is this:

```js
import { CableReadyElement } from "cable_ready"
CableReadyElement.define()
```

And that's it. The `cable-ready` tag will be loaded and registered, and then any time the tag gets inserted into any part of the page anywhere, boom, the operation will be performed and the tag immediately removed from the DOM.

Using CableCar is a breeze:

```ruby
CableReady::CableCar.instance
    .console_log(message: "Super duper!")
    .append("#root", html: "<strong>And a hearty hello to CableCar!</strong>")
    .dispatch(element: true)
```

that would output:

```html
<cable-ready><script type="application/json">[{"message":"Super duper!","operation":"consoleLog"},{"html":"\u003cstrong\u003eAnd a hearty hello to CableCar!\u003c/strong\u003e","selector":"#root","operation":"append"}]</script></cable-ready>
```

Let me know what you think!


leastbad:
That is a really clever idea and implementation!

Could you give some examples of what you'd use this for? What's the killer app? 😉 

Specifically, a couple of "I frequently use CR to X, which currently involves me doing Y, but this would allow me to do Z instead."

As you see with our efforts to make `cable_car` work effortlessly with `mrujs`, we're all for championing the CR standard format JSON across all sorts of really interesting domains. I just need a bit of help with the "why" on this one. 💙

jaredcwhite:
Let's say I'm getting an HTML partial back from the server I'll swap in somewhere in the DOM (using any library, vanilla JS, whatever). Contained within that HTML partial, I could add one or more CR operations. No need for any sort of special processing or handling. Since they're just web component tags, it means there doesn't need to be a certain request that's _only_ a CableCar dispatch and thus JS that knows to receive it and run `CableReady.perform`.

jaredcwhite:
I can work up a more elaborate scenario of what that might look like in practice if you think it's worthwhile.

jaredcwhite:
Maybe another way to put it is essentially it means you can have a "zero JS frontend￼" — not really because you need the one-liner to import and register the custom element – so in any context or codebase where you don't want to (or can't) change much of how the frontend works, you could still "sneak" in CR operations.

leastbad:
Zero-JS is actually (explicitly!) not a goal of SR/CR, but that doesn't detract from the potency of this PR/concept in my mind.

I feel like I get what this does, and how it does it, but I'm still a bit stuck on what I'd use it for. To be clear: this is a failure of my imagination, not a problem with the proposal. (I'm also a bit sleepy today, which doesn't help me see higher dimensions.)

Using web components to carry out an action and then self-destruct *feels* exciting and radical. I'm just trying to figure out what I'd be doing where the obvious solution is to embed operations in the page itself. For example, are you going to dispatch events? Remove DOM elements? Console log? Play sounds? I actually don't know!

Usually when people are learning SR and working with Page Morphs, they will often ask how to know when the Reflex is finished so their client-side JS can operate on the page. The answer is usually that if they render the page so that it's already done when it's delivered, they won't need to run any client-side JS. This doesn't always cover their problem case, but it does help them :bulb: on the mental model of serving finished HTML.

So that brings me back to `<cable-ready-operation operation-name="remove" selector="#foo">`. Specifically, if I don't want `#foo` in my DOM anymore... why not just remove it from the DOM before sending the HTML to the client?

jaredcwhite:
> I'm just trying to figure out what I'd be doing where the obvious solution is to embed operations in the page itself.

Hmm, trying to figure out a different way to explain this… This shouldn't be considered an *additional* solution beyond the current JSON-over-the-wire solution thus requiring a new set of use cases. It's quite possibly a 100% replacement for anyone who's interested. By sending operations as web components over the wire—like Turbo Streams—you can process them on the client side differently. In fact, to underscore this, you could easily use Turbo Streams to facilitate processing Cable Car operations. (wut)

```ruby
def some_controller_action
  render turbo_stream: turbo_stream.append("root", 
    cable_car
      .console_log(message: "Super duper!")
      .dispatch(elements: true)
  )
end
```

By allowing Cable Car to ride (heh) on top of an existing HTML-based transport mechanism (Turbo Streams), there's literally no extra setup of any kind required (other than the one-liner JS import for the custom element).

jaredcwhite:
Also updated the Bridgetown demo code so the operations are contained within a `template` tag and thus inert, and there's a button you can click which then appends the operations to the body. Which in turn calls the `confetti` custom operation which does a little poof of confetti on the page. And you can click it over and over again. 🤩 Custom operations are rad!

leastbad:
I appreciate the additional context (and comments on Discord) - helps a ton. I think I get it, now.

How would you feel about a plugin route? Just taking your temperature, here.

My current take is: the bigger the tent, the better. I think it's really powerful to tell TurboStreams folks that they have 35 new operations. (I wish that they'd just use cable_car, to be honest... it would be a transparent upgrade.)

We *do* need to make sure that people learning the stack aren't overwhelmed by the sheer number of comparable options, which is why I'm curious about the plugin approach. (We've already started extracting modules, like the `play_sound` operation...)

Others from the team will have plenty to say, I'm confident.

jaredcwhite:
> How would you feel about a plugin route?

Do you mean like a separate gem/NPM package? Possibly, but I'd want to make sure there's some kind of official way in so I don't have to monkey-patch stuff. Is there a desire to make CR more modular/plugin oriented?

leastbad:
I mean, we're working towards it! It's not something you can just flip a switch on, though.

It does sound like you have a lower estimation of monkey patching than I do, which is fine - a lot of folks don't like doing it. I'm just trying to figure out a way to make sure folks who want to use CR with TurboStreams can do so... while keeping all of the considerations in my last comment in mind.

Thinking out loud: what if we changed the `dispatch` function signature to `elements: nil` and then changed line 12 to:
```rb
payload = elements ? send(elements) : operations_payload
```

Then you can safely reopen `OperationBuilder` and add your `operations_custom_elements` method. Now you could do:

```rb
dispatch(elements: :operations_custom_elements)
```
(or a shorter method name!)

I recognize that this might not be as silky smooth but I'm looking for the BATNA.

jaredcwhite:
I think in the long run it's better for the community if this is a first-party concept, assuming there's interest. If I do set up a gem, I'd probably just define an `elements` method in lieu of `dispatch` entirely. The more I think about it, that's probably the only thing I'd have to touch within existing CR code.

leastbad:
I like your `elements` method idea, since it would mean no re-writing `dispatch`. However, I have a different suggestion in the same extended universe: `cable_ready` and `cable_car` both subclass `OperationBuilder`...

Perhaps what you're really looking for is a third Thing that emits HTML components instead of JSON? After all, the actual implementation of CableCar is like 15 lines soaking wet.

Still just thinking out loud.

ParamagicDev:
First off, I want to say this is pretty genius. When the element connects, it runs the CR operations and then removes itself. Pretty wild tbh. Secondly, the question I'm asking myself is where do we draw the line at Plugin vs Core?

I still dont have the proper gauge here as to whether this should live as a plugin or in core. I totally get both angles. Having a web component first class citizen opens up some cool opportunities especially for people streaming in DOM updates to the browser. 

I dont know how often Turbo Streams are combined with CR / SR....so perhaps this may be better as a plugin? Maybe we put this to a vote? Trust me, I totally get not wanting to have to setup a Gem + NPM package just to use this especially where its fairly minimal. 

I do worry about potential sneaky bugs with JSON embedded inside HTML, but thats a different story for a different day.

jaredcwhite:
However we decide to move forward on this, I'd love to make progress so let me know @leastbad what you'd need to know in order for me to put a bow on this PR, or alternatively, extract it out into a plugin. I still think having HTML-over-the-wire WCs as a first-party transport mechanism for CR is a compelling idea, but even as a separate gem/NPM package, it feels worth pursuing.

leastbad:
Sorry to leave you hanging! Waiting for feedback from the boss.

jaredcwhite:
I think I'll go ahead and close this so I can experiment with implementing it as a plugin. That way I can iterate faster towards a comprehensive solution and also try it out in some real-world scenarios. Always happy to talk about merging the plugin into "core" down the road.

hopsoft:
I really like the idea of encapsulating client side implementation details of DOM operations inside of custom elements. I can see a possible future where a foundational lib of custom elements exists that maps most/all standard DOM operations which could prove to be a low level building block for implementations like CableReady, TurboStreams, et al.

leastbad:
That's really interesting - you sort of "pivot tabled" Jared's proposal, right?

If I understand correctly, you could install a package and have access to a 1:1 set of custom elements (and matching Rails helpers) that correspond to CR operations?

jaredcwhite:
> I really like the idea of encapsulating client side implementation details of DOM operations inside of custom elements. I can see a possible future where a foundational lib of custom elements exists that maps most/all standard DOM operations which could prove to be a low level building block for implementations like CableReady, TurboStreams, et al.

So, that would mean instead of a single custom element which simply infers and passes JSON data along to `CableReady.perform`, there'd be multiple elements, each one of which would "wrap" a single operation? That does sound nifty…I'll have to think if there's a semi-automated way we could facilitate that. Maybe something that can run through the list of operations the JS library supports and "emit" an element for each operation.

Or are you saying the "current" JS library itself would switch gears—it would simply append these custom elements to the DOM and they would then handle the various CR operations internally? 🤔

leastbad:
> there'd be multiple elements, each one of which would "wrap" a single operation

hopsoft:
> there'd be multiple elements, each one of which would "wrap" a single operation

☝🏻 This option. A lot of DOM operations get pretty complex and I really like the idea of having a discreet element for each.

hopsoft:
Imagine a library of custom elements that encapsulate all the meaningful DOM operations that you may possibly want to trigger from the server. https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model

leastbad:
I know it's out of the blue, but I'm reopening this because in the year that's passed I've seen enough scenarios where a markup embedded solution would add value that I've fully bought in to what Jared was so patiently explaining.

Where I sort of lost the thread was when the scope started blowing up (love you, @hopsoft).

From my perspective, there's three main paths that this concept could take:

1. One generic web component that has an `operation` attribute and a Rails helper that would set it up. Call this the 1:1 operation:element approach, and I think it's what Jared proposes in this PR (I need to review).
2. One web component for each operation, as per Nate's vision. This is still a 1:1 approach, but instead of a generic component there's a component for each operation.
3. My proposal, which is a departure from the above, would be a single web component that represents the payload instead of the operation. This is literally just a container for encoded JSON that gets passed to `perform`.

My reasons for preferring door number three are:
- adding boilerplate and virtually identical shells to represent each operation seems like a lot of extra work, LOC and maintenance for questionable benefit (since most people will be using a helper to generate the markup)
- someone adding custom operations to CR would need to create web components, which really ramps up the currently straight-forward process of defining custom operations today
- some of my favourite features of CR 5 are related to the chaining of operations; indeed, the mental model of CR is that you can chain logical groups of operations together and they will be executed together, potentially in numbered batches

I'm picturing something like this:
```html
<cable-ready payload="[{\"message\":\"Super duper!\",\"operation\":\"consoleLog\"},{\"message\":\"Super duper!\",\"operation\":\"consoleLog\"},{\"message\":\"Super duper!\",\"operation\":\"consoleLog\"},{\"message\":\"Super duper!\",\"operation\":\"consoleLog\"},{\"elements\":true,\"operation\":\"dispatchEvent\"}]"/>
```

I'm sorry that the above isn't properly escaped, but you get the idea.

Could we rally around this?

jaredcwhite:
Excited to see this get picked up again @leastbad. I think your door number three is a good first step. For that payload concept, I might suggest a couple of alternatives to a literal `payload` attribute…either using `<template>` or `<script type="application/json">` as an inner tag to contain the JSON. For example:

```html
<cable-ready><script type="application/json">
[{"message":"Super duper!","operation":"consoleLog"},{"message":"Super duper!","operation":"consoleLog"},{"elements":true,"operation":"dispatchEvent"}]
</script></cable-ready>
```

I think `<template>` is more for holding HTML fragments than JSON, but OTOH it could be slightly less brittle in cases where the HTML is getting filtered by something which doesn't let `<script>` tags through. Thoughts?

leastbad:
I very much defer to your expertise on the web components front, Jared!

I agree that an [element body?] is likely superior to a `payload` attribute, even just on the basis that it's far easier to encode.

Given, again, that this element's HTML is almost certainly going to be generated by a helper, this might be somewhat moot... but between
```html
<cable-ready><script type="application/json">[{"message":"Super duper!","operation":"consoleLog"},{"message":"Super duper!","operation":"consoleLog"},{"elements":true,"operation":"dispatchEvent"}]</script></cable-ready>
```
and
```html
<cable-ready><template>[{"message":"Super duper!","operation":"consoleLog"},{"message":"Super duper!","operation":"consoleLog"},{"elements":true,"operation":"dispatchEvent"}]
</template></cable-ready>
```
why can't we just
```html
<cable-ready>[{"message":"Super duper!","operation":"consoleLog"},{"message":"Super duper!","operation":"consoleLog"},{"elements":true,"operation":"dispatchEvent"}]</cable-ready>
```
Pardon my ignorance! It just seems a bit ugly to nest an element, doesn't it?
Does wrapping it in a `script` element convey any advantage in terms of parsing it? Or is it functionally equivalent to a `template` in this scenario?

jaredcwhite:
Well you need the JSON payload not to be in a visible part of the DOM tree…otherwise it could show up in a FOUC scenario and mess up SEO and screen reading. The script tag is most ideal because the _only_ thing really you'd need to escape in the payload is a literal `</script>` string. For a template tag, you'd need to escape things in an XML-y fashion because the browser attempts to parse HTML tags and entities.

ParamagicDev:
Personally, I agree with the first approach of making the payload a JSON string.

<cable-ready operations="{}"></cable-ready>

It seems like its the safest way to embed the JSON in a declarative way without having to worry about DOM parsing or any other shenanigans.

jaredcwhite:
@ParamagicDev Yeah, you may be right and I'm overthinking it. 😄

leastbad:
I'm currently partial to the `<script type="application/json">` approach, because my understanding is that you could then just write un-escaped JSON into the body of that script tag. This is vastly less confusing to me than the hoops I seem to have to jump through to write JSON objects to attributes for Stimulus to process as values.

When you're writing an object to an attribute, suddenly you have to account for the fact that the attribute itself is opened and closed with a `"` character... forcing all `"` characters in the JSON to need escaping. That's miserable for people who are trying to author a tag by hand.

If we went with the `script` approach, we'have to test that the component element body contained a script tag, and then `JSON.parse()` it, but it should be fairly straight-forward, no?

leastbad:
Bump!

I really like where we arrived in our discussion. What's the best way forward?

Translated: Jared, this was and is your cool idea. Do you want to take a crack at morphing it into where we landed, or do you want to pass the torch?

jaredcwhite:
@leastbad Yeah I think that approach feels like the right solution for a v1. I'll work on it soon once I'm back on my feet.

jaredcwhite:
All right, I ended up realizing I could clean up a few things in a client project that used inelegant Turbo Stream stuff I'd cobbled together previously, so I decided to use that as a proving ground for this latest update. Very pleased with the results! (BTW, the previous `elements: true` in `dispatch` is now `element: true` because it's just a single element which gets rendered.)

In the process, I also realized I could do something pretty super-dee-dooper awesome. All you have to do is add this Rails initializer:

```ruby
module Turbo
  module Streams
    module CableCar
      def cable_car
        yield
        append_all "html", CableReady::CableCar.instance.dispatch(element: true)
      end
    end
  end
end

Turbo::Streams::TagBuilder.include Turbo::Streams::CableCar
```

And then…drum roll please 🥁…you can do this in a controller!

```ruby
# save stuff…

flash.now.notice = t("banks.published_notice")

render turbo_stream: turbo_stream.cable_car {
  cable_car.inner_html "#flashes", html: view_context.render("flashes")
  cable_car.outer_html "bank-publish-action", html: view_context.render("banks/publish_form", bank: @bank)
  cable_car.redirect_to url: bank_path(@bank), delay: 5000
}
```

Yes my friends, you can now ride the Cable Car on the Turbo Stream and it just works. Honestly I think this will be my canonical approach to rendering anything beyond a straight view from a controller going forward. 🙌

javascript/elements/cable_ready_element.js
```diff
@@ -0,0 +1,36 @@
+import { perform } from '../cable_ready'
+
+export default class CableReadyElement extends HTMLElement {
+  static define () {
+    if (!customElements.get('cable-ready')) {
```
leastbad, line 5: It seems like either this should be moved to https://github.com/stimulusreflex/cable_ready/blob/9fdbc8996d724473909ab708e1ad4cf25bad5386/javascript/elements/index.js#L28, or all of the other elements should be updated to use this `static define()` form. Which should it be?

leastbad (changes requested):
Great work so far, Jared!

I'm not a Turbo Streams user, but I still want the API to be as slick as possible. In https://github.com/stimulusreflex/cable_ready/pull/139#issuecomment-1134997748 you show `render turbo_stream: turbo_stream.cable_car {}`... what, if anything, can be done to get this down to `render turbo_stream: cable_car {}`?

Now, the big thing missing - which might speak to us having different intended use cases for this functionality - is the apparent absence of a `cable_ready_tag` helper in https://github.com/stimulusreflex/cable_ready/blob/master/app/helpers/cable_ready_helper.rb.

I would very much like to see a future where folks can do this:

```html
<%= cable_ready_tag cable_car.console_log(message: "hi") %>
```

In order for this to work, `cable_car` needs to be available to view helpers and the helper presumably needs to be able to call dispatch on what it receives as a parameter. Thoughts, please!

javascript/elements/cable_ready_element.js
jaredcwhite: So from what I understand, that particular function tries to set up CR connections and will throw an error if there's a missing consumer. I'd like the CR element itself to standalone, so a user needs to be able to import it by itself. Now whether we let them handle the `customElements.define` or we do it automatically upon import or we use a `define()` method, certainly up for debate.

jaredcwhite:
Re: `cable_ready_tag` — totally! I could work on that and it feels pretty simple to do.

On the Turbo front, I'm not sure if just having `cable_car` with a block would work because that means the `cable_car` helper method itself would need to assume it's being using within a Turbo scenario and output the whole Turbo Stream tag. Not to get all academic but it feels like that violates SRP. By hanging it right off of the `turbo_stream` helper, it's just another action you can take with `turbo_stream` which seems to make sense — thought I agree `render turbo_stream: turbo_stream.` is verbose. But so is Turbo's own `render turbo_stream: turbo_stream.append …` type of nomenclature.

javascript/elements/cable_ready_element.js
jaredcwhite: And in case my reasoning isn't super-clear, I personally would often use the CR element outside of Rails & ActionCable entirely, hence the need for maximum compatibility.

leastbad: I'd love to get @julianrubisch's and @marcoroth's :eyes: on this, but it seems like we should be consistent wrt the `define` technique. I'd be fine with this PR moving the defines for the other two elements into the elements themselves, as you have done for `cable-ready`.

As for the standalone question... I concede that this particular element does not need access to the `CableConsider` object which we rely on for the other two.

leastbad:
I think I hate the Turbo Stream API, then? But certainly nobody is asking me.

I guess, if it's consistent with the existing use cases for TS then one could even argue that providing a nicer API might actually break expectation.

Well, since I don't plan on using streams any time soon, I trust that you'll do everything possible to make it nice for those that do; no violations of least surprise.

If we can nail a really nice API mouthfeel for `cable_ready_tag` - I think that's what it should be, right? - then I'll be a happy camper.

jaredcwhite:
@leastbad Seems logical to me. I've added that and a test which also tests the validity of the custom element output. I'm not sure what to do for documentation but I know that's your wheelhouse so just let me know. Excited to see this reach the home stretch!

hopsoft:
> I'm not sure if just having `cable_car` with a block would work because that means the `cable_car` helper method itself would need to assume it's being using within a Turbo scenario

This is partly _tongue in cheek_, but what about a `turbo_cable_car` helper. That would look something like this.

```ruby
render turbo_stream: turbo_cable_car { ... }
```

jaredcwhite:
@hopsoft 😄

Perhaps a bit goofy at first glance…but maybe that's the direction to go. Certainly memorable!

jaredcwhite:
@hopsoft @leastbad Any additional thoughts on if we should actually add that…either in this PR or a followup?

julianrubisch:
Just wanted to chime in and say how much I love the spirit of this 🤩

Only thing I want to point out is that we should take care to make this element as extensible as possible, since it could provide the basis for a much much broader API if we do it right.

leastbad:
> @hopsoft @leastbad Any additional thoughts on if we should actually add that…either in this PR or a followup?

Nate has final say, but I was more than convinced by your earlier comment that the Turbo Streams API itself uses the longer form; that we should mimic their form by leaving it as it is.

We can always add more *stuff* later.

javascript/elements/cable_ready_element.js
marcoroth: > So from what I understand, that particular function tries to set up CR connections and will throw an error if there's a missing consumer

@jaredcwhite This function is the initialize function to setup the whole library, not just the consumer. It's usually being call in the app code in `application.js`. So it would make sense to keep the `define` calls here.

But maybe we can move the define logic into the elements itself, so that the elements can handle themselves how they are being declared. This also has the benefit that the name of the element is at the same place as the code for the element itself.. And instead of defining it directly in the `initialize()` function we just call the static `declare()` function on every element.

Something like:
```js
// javascript/elements/index.js

import StreamFromElement from './stream_from_element'
import UpdatesForElement from './updates_for_element'
import CableReadyElement from './cable_ready_element'


// ...

const initialize = (initializeOptions = {}) => {
 // ..

 StreamFromElement.declare()
 UpdatesForElement.declare()
 CableReadyElement.declare()

 // ..
}
```

jaredcwhite: @marcoroth @leastbad OK I wrapped my head around this, so I've updated the PR with the consolidated logic for defining the elements.

jaredcwhite:
@hopsoft @leastbad Just thought I'd check in to see if there's a sense of when this might get reviewed. I'd love to help it head into the final stretch.

hopsoft:
I'm planning on getting to this within the next week. Thanks for being so patient.

hopsoft (changes requested):
I'd love to get this merged soon. I think the only thing missing here is the turbo stream render helper. How difficult will be to get something like this added to the PR?

```ruby
render turbo_stream: cable_ready_operations { ... }
```

hopsoft:
@jaredcwhite Can you share a real world snippet of how you're doing #2 in your project(s)?

hopsoft:
I'd really love an API like this, but am not sure how practical the implementation might be as I haven't dug into it yet.

```ruby
render turbo_stream: operations {
  inner_html "#flashes", html: view_context.render("flashes")
  outer_html "bank-publish-action", html: view_context.render("banks/publish_form", bank: @bank)
  redirect_to url: bank_path(@bank), delay: 5000
}
```

jaredcwhite:
> @jaredcwhite Can you share a real world snippet of how you're doing #2 in your project(s)? Assuming [this](https://github.com/stimulusreflex/cable_ready/pull/139#issuecomment-1134997748) _(or something like it)_ is what you're doing?

@hopsoft Yeah, that's basically it.

I love the look of the simplified API example, but practically I'm not sure how that's possible without a lot of Ruby-fu under the hood (aka use `instance_exec`, delegate all `method_missing` calls to the underlying controller, etc.)

jaredcwhite:
The name of what gets passed to `turbo_stream:` is also critical: in that simplified example, `operations` means that the Rails controller itself has a method called `operations`. In the example of using `turbo_stream: turbo_stream.cable_car` it means that the `turbo_stream` controller (and view) helper itself has a new method of `cable_car` which can be used anywhere that helper is available.

hopsoft:
I'm good sticking with `turbo_stream: turbo_stream.cable_car` for now unless anyone has objections.

hopsoft:
That effectively gives us something like this.

```ruby
render turbo_stream: turbo_stream.cable_car {
  cable_car
    .inner_html("#flashes", html: view_context.render("flashes"))
    .outer_html("bank-publish-action", html: view_context.render("banks/publish_form", bank: @bank))
    .redirect_to(url: bank_path(@bank), delay: 5000)
}
```

jaredcwhite:
Yep. OK so I can work on adding that implementation in early this week. Sounds like we're probably good otherwise?

leastbad:
Earlier in the conversation, Jared mentioned that the syntax for working with TS was
```ruby
render turbo_stream: turbo_stream.append "messages", @message
```
That syntax is not beautiful - the repetition of `turbo_stream` is jarring to my eyes - but it's reasonable. You're rendering a turbo stream, and filling it with a turbo_stream.append.
```ruby
render turbo_stream: turbo_stream.cable_car {
  cable_car.append("#messages", @message)
}
```
We're replicating `turbo_stream` *and* `cable_car`.

Is it impossible for us to get to something like
```ruby
render turbo_stream: turbo_stream.cable_car.append("#messages", @message)
```
???

We have always really put a lot of care into creating a beautiful API surface. It's worth it to get this right!

jaredcwhite:
Meanwhile this has just landed, shockingly!

https://github.com/hotwired/turbo/pull/479
https://github.com/hotwired/turbo-site/pull/107/files (docs)

So…hmm, it's tempting to wonder if we could create a Turbo Stream action extension directly that can process CR operations inside its template. I think we can continue this PR as-is with the custom element for general use cases, but in terms of a Turbo extension specifically, maybe we actually go for it in a more canonical manner now. It would require some extra JS though, and not just extending on the Ruby side.

hopsoft:
Agree with this assessment. 🚀

hopsoft (approved):

jaredcwhite:
I agree with shipping this as-is and leaving Turbo integration as a separate consideration.

leastbad:
Thanks for all of your hard work on this, @jaredcwhite! Quite the journey. 😀

PR #146, opened 2021-08-31
Fix wrong CableReady import

n-rodriguez:
Fix this error:

```sh
ERROR in ./.yarn/cache/cable_ready-npm-5.0.0-pre3-b6a471a50a-bf6c2acdbd.zip/node_modules/cable_ready/javascript/stream_from_element.js 15:33-51
export 'default' (imported as 'CableReady') was not found in 'cable_ready' (possible exports: initialize, perform, performAsync)
```

### DIFF
--- a/javascript/stream_from_element.js
+++ b/javascript/stream_from_element.js
@@ -1,4 +1,4 @@
-import CableReady from 'cable_ready'
+import { perform } from './cable_ready'
 import { consumer } from './action_cable'
 
 class StreamFromElement extends HTMLElement {
@@ -12,7 +12,7 @@ class StreamFromElement extends HTMLElement {
         },
         {
           received (data) {
-            if (data.cableReady) CableReady.perform(data.operations)
+            if (data.cableReady) perform(data.operations)
           }
         }
       )


javascript/stream_from_element.js
```diff
@@ -1,4 +1,4 @@
-import CableReady from 'cable_ready'
+import CableReady from './index'
```
marcoroth, line 1:
```suggestion
import { perform } from './cable_ready'
```

marcoroth (commented):
Hey @n-rodriguez

Thank for this pull request!

I think it would even make more sense if we just import the `peform` function from the `cable_ready.js` file and update [L15](https://github.com/stimulusreflex/cable_ready/blob/master/javascript/stream_from_element.js#L15) to:

```js
            if (data.cableReady) perform(data.operations)
```

What do you think?

n-rodriguez:
@marcoroth good catch! it's fixed.

marcoroth (approved):

marcoroth:
Thank you @n-rodriguez!

leastbad:
Great work on this.

PR #154, opened 2021-10-05
post-pre4 updates

leastbad:
- removed sideEffects from package.json
- renamed Boris function to debouncer
- refactored updates_for to support different URLs
- adds and removes `updating` attribute for CSS

Outstanding points and questions:
- I'm a little fuzzy on how we should be handling `assignFocus` with multiple blocks being updated
- right now, debounce is set by the first web component; specifying a different value on a later `updates_for` will have no effect
- currently, the `updating` attribute is added and removed to and from each element in a cascade; if you want all elements to receive and lose their `updating` attribute at the same time, I will have to rework this slightly
- I'm not 100% sure what happens if there aren't enough blocks on a page referenced by `url`, this might require a try/catch

"Test" case:
```erb
<!-- home/index -->
<%= updates_for current_user do %>
  <p><%= rand(1..1000) %></p>
<% end %>
<%= updates_for current_user, url: "home/b" do %>
  <p><%= rand(1..1000) %></p>
<% end %>
<%= updates_for current_user, url: "home/a" do %>
  <p><%= rand(1..1000) %></p>
<% end %>
<%= updates_for current_user, url: "home/b" do %>
  <p><%= rand(1..1000) %></p>
<% end %>
<%= updates_for current_user, url: "home/a" do %>
  <p><%= rand(1..1000) %></p>
<% end %>
```
```erb
<!-- home/a -->
<%= updates_for current_user do %>
  <p>a1</p>
<% end %>
<%= updates_for current_user do %>
  <p>a2</p>
<% end %>
<%= updates_for current_user do %>
  <p>a3</p>
<% end %>
<%= updates_for current_user do %>
  <p>a4</p>
<% end %>
<%= updates_for current_user do %>
  <p>a5</p>
<% end %>
```
```erb
<!-- home/b -->
<%= updates_for current_user do %>
  <p>b1</p>
<% end %>
<%= updates_for current_user do %>
  <p>b2</p>
<% end %>
<%= updates_for current_user do %>
  <p>b3</p>
<% end %>
<%= updates_for current_user do %>
  <p>b4</p>
<% end %>
<%= updates_for current_user do %>
  <p>b5</p>
<% end %>
```
When the above scenario is triggered, it results in:
```
412
b2
a3
b4
a5
```

### DIFF
--- a/javascript/updates_for_element.js
+++ b/javascript/updates_for_element.js
@@ -22,11 +22,12 @@ class UpdatesForElement extends HTMLElement {
     super()
     const shadowRoot = this.attachShadow({ mode: 'open' })
     shadowRoot.innerHTML = template
-    this.update = debounce(this.update.bind(this), this.debounce)
   }
 
   async connectedCallback () {
     if (this.preview) return
+    this.update = debounce(this.update.bind(this), this.debounce)
+
     const consumer = await actionCable.getConsumer()
     if (consumer) {
       this.channel = consumer.subscriptions.create(
@@ -78,11 +79,13 @@ class UpdatesForElement extends HTMLElement {
       dispatch(blocks[i], 'cable-ready:before-update', operation)
       morphdom(blocks[i], fragments[i], {
         childrenOnly: true,
-        onBeforeElUpdated: shouldMorph(operation)
+        onBeforeElUpdated: shouldMorph(operation),
+        onElUpdated: _ => {
+          blocks[i].removeAttribute('updating')
+          dispatch(blocks[i], 'cable-ready:after-update', operation)
+          assignFocus(operation.focusSelector)
+        }
       })
-      blocks[i].removeAttribute('updating')
-      dispatch(blocks[i], 'cable-ready:after-update', operation)
-      assignFocus(operation.focusSelector)
     }
   }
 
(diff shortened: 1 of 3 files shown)


javascript/updates_for_element.js
```diff
@@ -13,12 +13,16 @@ const template = `
...
+  return ele.hasAttribute('url') ? ele.getAttribute('url') : location.href
+}
+
 class UpdatesForElement extends HTMLElement {
   constructor () {
     super()
     const shadowRoot = this.attachShadow({ mode: 'open' })
     shadowRoot.innerHTML = template
-    this.update = Boris(this.update.bind(this), this.debounce)
+    this.update = debounce(this.update.bind(this), this.debounce)
```
julianrubisch, line 25: I just noticed a regression here, but this has probably already crept up during the last refactoring:

There's no chance the element can have a `debounce` attribute at the time of construction, i.e. `this.debounce` will always return 20. So in my case I'm seeing thundering herds again.

I think we need to move this to `connectedCallback`

package.json
```diff
@@ -25,7 +25,6 @@
   "author": "Nathan Hopkins <natehop@gmail.com>",
   "main": "./javascript/index.js",
   "module": "./javascript/index.js",
-  "sideEffects": false,
```
julianrubisch, line 28: apparently this broke elements loading? not hitting the constructor anymore...

javascript/updates_for_element.js
leastbad: I think that you're confusing the `debounce` function that is being imported with the `this.debounce` getter available inside the component.

julianrubisch:
Can we merge this?

PR #156, opened 2021-10-06
Export elements

julianrubisch:
This PR moves defining the elements to `CableReady.initialize`. Since this has to be called anyway, I figured this might be the best place to do it.

Also, it exports `StreamFromElement` and `UpdatesForElement` for outside use.


leastbad:
Actual LOL - I'm so glad that I let you do this.

LGTM!

PR #158, opened 2021-10-08
Re-export consumer

julianrubisch:
As a further simplification, re-export the consumer so it can easily be obtained by calling

`await CableReady.consumer`

This can even be re-used among sub-libraries like StimulusReflex, Futurism, and potentially more


leastbad:
This is great. Good work.

PR #163, opened 2021-11-12
updates_for :only option

### LINKED ISSUE #161: updates_for: Support multiple enable_updates per model
Hi there! I've had a lot of fun playing with `updates_for` since it went live so many thanks for sharing it with the world.

Just wondering whether it might make sense to allow for multiple `enable_updates` on a given model, which then map to different views.

For example, let's say I have a `Blog` model with a show page and an index page. And let's just pretend these blogs get updated quite frequently by their authors.

On the index page I only need to update the name of a given blog if it's changed, so I might do something like this:

```rb
class Blog < ApplicationRecord
  enable_updates on: :update, if: -> { saved_change_to_name? }
end
```

Now I start building the show page and want to enable updates if _any_ field changes. I could go back to my `enable_updates` and remove the `if` Proc, but then my index page blogs would be constantly fetching updates even when none of the fields changed are visible on that page.

One thing I love about `updates_for` is its simplicity, so I'm hesitant to propose anything that takes away from that. That said, I also feel like allowing for more granular setting of when/where to enable updates would have widespread application beyond something like the contrived example above.

Penny for your thoughts :)

leastbad:
Closes #161

While I believe that a Model has no business knowing anything about views, @andrewerlanger's desire to eliminate unnecessary HTTP requests is highly reasonable. Therefore, instead of attempting to further complicate the `enable_updates` method with the potential for multiple calls that each target different views, this PR gives the `updates_for` helper the ability to optionally filter out messages which don't contain the right keys.

```js
<%= updates_for User.first, only: "name" do %>
  <p><%= rand(1..1000) %></p>
<% end %>
```

```rb
User.first.update name: "Bob" # triggers refresh
User.first.update age: 43, name: "Steve" # triggers refresh
User.first.update age: 43 # does not trigger refresh
```

Any broadcasts which update individual model instances now transmit an array of attribute names that were updated:

```rb
ActionCable.server.broadcast(identifier, {changed: model.previous_changes.keys})
```

The `only` option currently accepts only a simple string that exactly matches the name of an attribute. There's no checking to ensure that it is a valid attribute, and there's no support for `and`/`or`-ing multiple attributes. I don't think that we need this feature to be Turing complete to be Just The Right Amount of functionality.

Note that this feature has no impact on Model messages (create/destroy) because it only works on `after_update` callbacks for individual model instances.


app/models/concerns/cable_ready/updatable.rb
```diff
@@ -43,9 +43,9 @@ def cable_ready_collections
         @cable_ready_collections ||= CollectionsRegistry.new
       end
 
-      def cable_ready_update_collection(resource, name)
+      def cable_ready_update_collection(resource, name, model)
         identifier = resource.to_global_id.to_s + ":" + name.to_s
-        ActionCable.server.broadcast(identifier, {})
+        ActionCable.server.broadcast(identifier, {changed: model.previous_changes.keys})
```
julianrubisch, line 48: I think this will need a `respond_to? :previous_changes` guard to account for POROs, as exemplified by the need to add that method to the test dummy PORO

app/models/concerns/cable_ready/updatable/model_updatable_callbacks.rb
```diff
@@ -21,7 +21,7 @@ def broadcast_create(model)
 
       def broadcast_update(model)
         ActionCable.server.broadcast(model.class, {})
-        ActionCable.server.broadcast(model.to_global_id, {})
+        ActionCable.server.broadcast(model.to_global_id, {changed: model.previous_changes.keys})
```
julianrubisch, line 24: See above

julianrubisch (changes requested):

app/models/concerns/cable_ready/updatable.rb
leastbad: TIL

javascript/elements/updates_for_element.js
```diff
@@ -40,12 +40,15 @@ export default class UpdatesForElement extends SubscribingElement {
...
 
-  async update () {
+  async update (data) {
     const identifier = this.getAttribute('identifier')
     const query = `updates-for[identifier="${identifier}"]`
     const blocks = document.querySelectorAll(query)
     if (blocks[0] !== this) return
 
+    const only = this.getAttribute('only')
+    if (only && data.changed && !data.changed.includes(only)) return
```
andrewerlanger, line 50: @leastbad just out of interest, would we be able to tweak this to allow for passing multiple fields into the `only` option?

I was thinking of something along the lines of:

```js
const only = this.getAttribute('only')?.split(" ")
if (only && data.changed && !only.some(attribute => data.changed.includes(attribute))) return
```

Which, if I understand correctly, would allow you to do:

```erb
<%= updates_for User.first, only: [:name, :email, :favorite_abba_song] do %>
  <p><%= rand(1..1000) %></p>
<% end %>
```

leastbad: `some` would make it an `OR`, right?

Just to play the devil's advocate, can you make a strong case why we'd go with `OR` over `AND` (or `NOR`/`XOR` 😉)

leastbad: BTW, this isn't just a pedantic question... technically, `AND` would allow developers to build up simple logical constraints eg "this one needs name and age, that one needs name and gender"

Whereas `OR` is probably better for an admin scenario where you have a bunch of user-editable data alongside a busy attribute such as a counter cache that you seek to filter out.

leastbad: What do you think, @julianrubisch?

julianrubisch: Just reading up on it here...

Very valid question. For me it boils down to the principle of least surprise:

If in Rails I do `before_action only: %i[new edit]` I expect it to work on both, right? So that'd make a case for an `OR`.

Can anyone think of a counter-example?

andrewerlanger: @leastbad yes, `some` would make it an `OR`.

I hadn't really considered the case for `AND` to be honest. And this is probably just a lack of imagination on my part, but I can't picture an obvious application for it.

I was operating on the assumption that you'd usually want `only` to include all attributes that feature within the `updates_for` block.

```erb
<%= updates_for User.first, only: [:name, :age] do |user| %>
  <p><%= user.name %></p>
  <p><%= user.age %></p>
<% end %>

<%= updates_for User.first, only: [:name, :gender] do |user| %>
  <p><%= user.name %></p>
  <p><%= user.gender %></p>
<% end %>
```

In cases like this, I imagine you'd usually want to update the contents of the block any time _any_ supplied attribute is updated, to keep the HTML content in-sync with the database.

I'm sure there are some cases in which `AND` would be the better fit, I just have the sense they'd be a lot more rare. Also interested in @julianrubisch's thoughts on this :)

leastbad: I think both of you made the right points.

@andrewerlanger, do you feel up to making a PR against this PR to implement the changes to both `updates_for_element.js` and `cable_ready_helper.rb`?

It's important that the helper accept a string, a symbol and an array of strings or symbols, and of course, the web component will have to be able to handle whatever shows up.

I actually don't know how `tag` will serialize an array. Will it do space-separated tokens? comma-separated?

andrewerlanger: @leastbad happy to take a stab at it!

I'm not sure we'll actually need to make any changes to `cable_ready_helper.rb` though. As far as I can tell, any array we pass to `only` ends up becoming a (space-separated) string by the time it makes its way to the javascript.

I just added this to the `updates_for_element.js` code...

```js
const only = this.getAttribute('only')?.split(" ")
console.log(only)
```

...and got the following:
```js
// only: [:name, :age]
['name', 'age']

// only: ["name", "age"]
['name', 'age']

// only: "name age"
['name', 'age']

// only: "name"
['name']

// only: :name
['name']
```

So I think we should be in the clear on that front, but let me know if I've missed something.

I'll take a shot at those `updates_for_element.js` changes soon :)

leastbad: Nice work! I had to make a quick tweak, but it's now in there and seems to be working great.

PR #164, opened 2021-11-18
Query all eager loading turbo frames and patch their innerHTML

julianrubisch:
Allows to fetch and patch lazy loaded turbo frames present in the response eagerly.

### DIFF
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -61,43 +61,15 @@ export default class UpdatesForElement extends SubscribingElement {
       blocks[i].setAttribute('updating', 'updating')
 
       if (!html.hasOwnProperty(url(blocks[i]))) {
-        html[url(blocks[i])] = await graciouslyFetch(url(blocks[i]), {
+        const response = await graciouslyFetch(url(blocks[i]), {
           'X-Cable-Ready': 'update'
         })
+        html[url(blocks[i])] = await response.text()
       }
 
       template.innerHTML = String(html[url(blocks[i])]).trim()
 
-      const reloadingTurboFrames = [
-        ...template.content.querySelectorAll(
-          'turbo-frame[src]:not([loading="lazy"])'
-        )
-      ]
-
-      await Promise.all(
-        reloadingTurboFrames.map(frame => {
-          return new Promise(async resolve => {
-            const frameResponse = await graciouslyFetch(
-              frame.getAttribute('src'),
-              {
-                'Turbo-Frame': frame.id
-              }
-            )
-
-            const frameTemplate = document.createElement('template')
-            frameTemplate.innerHTML = await frameResponse.text()
-
-            template.content.querySelector(
-              `turbo-frame#${frame.id}`
-            ).innerHTML = String(
-              frameTemplate.content.querySelector(`turbo-frame#${frame.id}`)
-                .innerHTML
-            ).trim()
-
-            resolve()
-          })
-        })
-      )
+      await this.resolveTurboFrames(template.content)
 
       const fragments = template.content.querySelectorAll(query)
 
@@ -123,6 +95,43 @@ export default class UpdatesForElement extends SubscribingElement {
         }
       })
     }
+  }
+
+  async resolveTurboFrames (documentFragment) {
+    const reloadingTurboFrames = [
+      ...documentFragment.querySelectorAll(
+        'turbo-frame[src]:not([loading="lazy"])'
+      )
+    ]
+
+    return Promise.all(
+      reloadingTurboFrames.map(frame => {
+        return new Promise(async resolve => {
+          const frameResponse = await graciouslyFetch(
+            frame.getAttribute('src'),
+            {
+              'Turbo-Frame': frame.id,
+              'X-Cable-Ready': 'update'
+            }
+          )
+
+          const frameTemplate = document.createElement('template')
+          frameTemplate.innerHTML = await frameResponse.text()
+
+          // recurse here to get all nested eager loaded frames
+          await this.resolveTurboFrames(frameTemplate.content)
+
+          documentFragment.querySelector(
+            `turbo-frame#${frame.id}`
+          ).innerHTML = String(
+            frameTemplate.content.querySelector(`turbo-frame#${frame.id}`)
+              .innerHTML
+          ).trim()
+
+          resolve()
+        })
+      })
+    )
   }
 
   get debounce () {
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -4,7 +4,7 @@ import CableReady from '..'
 import SubscribingElement from './subscribing_element'
 import { shouldMorph } from '../morph_callbacks'
 import activeElement from '../active_element'
-import { debounce, assignFocus, dispatch, handleErrors } from '../utils'
+import { debounce, assignFocus, dispatch, graciouslyFetch } from '../utils'
 
 const template = `
 <style>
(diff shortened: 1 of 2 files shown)


javascript/elements/updates_for_element.js
```diff
@@ -73,6 +73,35 @@ export default class UpdatesForElement extends SubscribingElement {
...
+        ...template.content.querySelectorAll(
+          'turbo-frame[src]:not([loading="lazy"])'
+        )
+      ]
+
+      await Promise.all(
+        reloadingTurboFrames.map(frame => {
+          return new Promise(async resolve => {
+            const frameResponse = await fetch(frame.getAttribute('src'), {
+              headers: { 'Turbo-Frame': frame.id }
```
leastbad, line 87: Tangent: what's the deal with `X-Cable-Ready` on line 66? It doesn't occur anywhere else in the CableReady codebase.

javascript/elements/updates_for_element.js
```diff
@@ -73,6 +73,35 @@ export default class UpdatesForElement extends SubscribingElement {
...
+      const reloadingTurboFrames = [
+        ...template.content.querySelectorAll(
+          'turbo-frame[src]:not([loading="lazy"])'
+        )
+      ]
+
+      await Promise.all(
+        reloadingTurboFrames.map(frame => {
+          return new Promise(async resolve => {
+            const frameResponse = await fetch(frame.getAttribute('src'), {
```
leastbad, line 86: It seems like this `fetch` should have the ability to report an error, similar to https://github.com/stimulusreflex/cable_ready/blob/d3e2efcffaa0646077e317cd9b6f4c3b27b45524/javascript/elements/updates_for_element.js#L70

javascript/elements/updates_for_element.js
```diff
@@ -73,6 +73,35 @@ export default class UpdatesForElement extends SubscribingElement {
...
       template.innerHTML = String(html[url(blocks[i])]).trim()
+
+      const reloadingTurboFrames = [
+        ...template.content.querySelectorAll(
+          'turbo-frame[src]:not([loading="lazy"])'
+        )
+      ]
+
+      await Promise.all(
+        reloadingTurboFrames.map(frame => {
```
leastbad, line 84: Is there a nightmare scenario where a lazy-loaded frame could contain additional frames that themselves might be lazy-loaded?

It seems like perhaps this should actually be a call to a recursive-capable function, if so.

leastbad (commented):
So far as I understand the functionality of lazy-loaded Turbo Frames, this looks sound to me. I guess they use DOM ids to identify frames? For some reason, I thought they gave them arbitrary identifiers.

Depending on your reaction to one of my questions, you might need to make the lazy fetch into a recursive function.

julianrubisch:
Yep, frames all use IDs right now.

javascript/elements/updates_for_element.js
```diff
@@ -73,6 +73,35 @@ export default class UpdatesForElement extends SubscribingElement {
       }
 
       template.innerHTML = String(html[url(blocks[i])]).trim()
+
+      const reloadingTurboFrames = [
+        ...template.content.querySelectorAll(
+          'turbo-frame[src]:not([loading="lazy"])'
```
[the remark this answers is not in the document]
julianrubisch: Yeah there has been some confusion for me, too, but I think what we're doing is correct:

https://turbo.hotwired.dev/reference/frames#eager-loaded-frame

I've adopted the same terminology for futurism and also updated the PR title.

It also seems congruent with the experimental behavior on : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading

javascript/elements/updates_for_element.js
```diff
@@ -73,6 +73,35 @@ export default class UpdatesForElement extends SubscribingElement {
...
+        )
+      ]
+
+      await Promise.all(
```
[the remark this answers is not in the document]
julianrubisch: @ParamagicDev ?

javascript/elements/updates_for_element.js
julianrubisch: cf https://github.com/stimulusreflex/cable_ready/pull/159

To do something like
`before_action { layout false if request.headers['X-Cable-Ready'] == 'update' }` etc etc

javascript/elements/updates_for_element.js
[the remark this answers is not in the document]
leastbad: I mean, if it works in your test environment...! I don't use Frames so it's hard for me to test.

It just struck me as odd.

javascript/elements/updates_for_element.js
julianrubisch: I'm definitely considering this, just need a good test bench for this

javascript/elements/updates_for_element.js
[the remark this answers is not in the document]
ParamagicDev: Yes. If theres no promises to await, it just carries on

javascript/elements/updates_for_element.js
```diff
@@ -61,15 +61,20 @@ export default class UpdatesForElement extends SubscribingElement {
...
-          .catch(e => console.error(`Could not fetch ${url(blocks[i])}`))
-        if (response === undefined) return
-        html[url(blocks[i])] = await response.text()
+        try {
+          const response = await fetch(url(blocks[i]), {
+            headers: {
+              'X-Cable-Ready': 'update'
+            }
+          })
+          if (response === undefined) return
```
ParamagicDev, line 70: `response == undefined` or `response == null` is probably preferred here! 

```js
null === undefined // False!
null == undefined // True!
``` 

Not a huge deal, but one of those fun quirks to be in mind. Granted, this will raise an error once it gets to the next line for `handleErrors`.

julianrubisch: Ah, nice one! I'm thinking about how to best extract that whole functionality to a custom `fetch` routine

leastbad:
Excited to see this coming together.

julianrubisch:
So I can declare success! I threw together a couple of test cases including nested eager loaded frames, and it worked.

Thanks to @ParamagicDev for your support and reviewing it en passent 🙌

javascript/utils.js
```diff
@@ -114,6 +114,29 @@ function handleErrors (response) {
...
+// A proxy method to wrap a fetch call in error handling
+//
+// * url - the URL to fetch
+// * additionalHeaders - an object of additional headers passed to fetch
+//
+async function graciouslyFetch (url, additionalHeaders) {
+  try {
+    const response = await fetch(url, {
+      headers: {
+        'X-REQUESTED-WITH': 'XmlHttpRequest',
```
leastbad, line 126: Have we reached the `Mozilla/5.0` era of fetch request headers?

javascript/utils.js
```diff
@@ -114,6 +114,29 @@ function handleErrors (response) {
   return response
 }
 
+// A proxy method to wrap a fetch call in error handling
+//
+// * url - the URL to fetch
+// * additionalHeaders - an object of additional headers passed to fetch
+//
+async function graciouslyFetch (url, additionalHeaders) {
```
leastbad, line 122: Just curious, why "graciously"?

Why not `recursiveFetch` or `deepFetch` or `fetchWithPrejudice`?

javascript/elements/updates_for_element.js
```diff
@@ -99,6 +97,43 @@ export default class UpdatesForElement extends SubscribingElement {
     }
   }
 
+  async resolveTurboFrames (documentFragment) {
+    const reloadingTurboFrames = [
+      ...documentFragment.querySelectorAll(
+        'turbo-frame[src]:not([loading="lazy"])'
```
leastbad, line 103: I'm still puzzled that Turbo Frames puts `loading="lazy"` only on non-lazy instances. What am I not understanding? I trust that it works, it's just so bizarre.

leastbad (approved):
LGTM with comments re: naming and a request for clarification.

javascript/utils.js
julianrubisch: @ParamagicDev recommended it to be able to do `request.xhr?` in Rails

javascript/utils.js
julianrubisch: because we handle errors graciously. It's neither recursive nor deep, it just wraps `fetch` complete with error handling.

javascript/elements/updates_for_element.js
julianrubisch: I know it's blasphemous to ask this from _you_, but refer to the docs: 

https://turbo.hotwired.dev/reference/frames#eager-loaded-frame
https://turbo.hotwired.dev/reference/frames#lazy-loaded-frame

the distinction between them is that the latter uses an IntersectionObserver, whereas the former doesn't.

So my reasoning was, when something shouldn't be loaded until it's visible, it's probably best to follow along in this behavior.

leastbad: That's a great answer! Thank you.

I admit that my eyes glazed over the distinction.

PR #166, opened 2021-11-25
Draft: Futurism updates-for support

xtr3me:
This PR accompanies the following Futurism PR: https://github.com/stimulusreflex/futurism/pull/120

Since I found no other way to instruct morphdom to not mutate the futurism-element I introduced a data attribute to instruct the updates-for element to not call morphdom in the first place.

Secondly I introduced a data attribute that is used to emit a event to the provided selector. This allows futurism to listen to the event and call it's futurism actioncable channel and refresh it's partial without glitching

### DIFF
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -84,16 +84,26 @@ export default class UpdatesForElement extends SubscribingElement {
         html: fragments[i],
         permanentAttributeName: 'data-ignore-updates'
       }
-      dispatch(blocks[i], 'cable-ready:before-update', operation)
-      morphdom(blocks[i], fragments[i], {
-        childrenOnly: true,
-        onBeforeElUpdated: shouldMorph(operation),
-        onElUpdated: _ => {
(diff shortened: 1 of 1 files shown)


javascript/elements/updates_for_element.js
```diff
@@ -84,16 +84,26 @@ export default class UpdatesForElement extends SubscribingElement {
...
+          onBeforeElUpdated: shouldMorph(operation),
+          onElUpdated: _ => {
+            blocks[i].removeAttribute('updating')
+            dispatch(blocks[i], 'cable-ready:after-update', operation)
+            assignFocus(operation.focusSelector)
+          }
+        })
+      }
+
+      if (blocks[i].dataset.afterUpdateEventSelector) {
```
julianrubisch, line 101: I think this should happen inside the `onElUpdated` too?

julianrubisch:
I'm closing this as a lot of time has since passed and the implementations diverged a bit. Feel free to message me if you still sense an enhancement is due here. We're cleaning up PRs for the 5.0 release.

Issue #167, opened 2021-11-25
updates_for: skip updates


assuntaw:
Hey everyone! Big thanks for `updates_for` – it's been an absolute joy to work with 🙌  

One issue that I've come across is that, because `updates_for` requires an open Redis connection, CRUDing records will fail where you wouldn’t ordinarily have started a Redis server: seeds and data migrations (using `data-migrate`) in my case.

Here’s an example: 

```ruby
# app/models/chapter.rb
class Chapter < ActiveRecord
  include CableReady::Updatable

  enable_updates
end 

# db/seeds.rb
puts “Destroying chapters…”
Chapter.destroy_all

puts “Creating chapters…”
Chapter.create!
# …
```

```bash
bin/rails db:seed

Destroying chapters…
rails aborted!
Redis::CannotConnectError: Error connecting to Redis on localhost:6379 (Errno::ECONNREFUSED)
```

Spinning up a Redis server to run seeds has been ok (though not desirable) for me, but not being able to run (and, more importantly, roll back) data migrations without one seems a little more ✨ wild ✨.

I’m wondering whether it might make sense to add an option a la “skip updates”. Here are two ideas from an initial brainstorm: 

#### 1. A global config option that skips the relevant callbacks for predefined paths: 

```ruby
config.skip_updates_from += [Rails.root.join(“/db”)]
```

I’d personally like this from a DX perspective as I can’t see a need to broadcast changes from anywhere in the `/db` directory, so a blanket exclusion would make sense.

I’m not 100% sure what a reliable implementation would look like though – intuition tells me that foraging through the callstack in an `after_commit` conditional could be very error-prone. It’s also a pretty rigid approach (though maybe that’s what you’d want here?).

#### 2. A `skip_updates` method that yields a given block (credit to @andrewerlanger for this idea, which I also ~like~ love):

```ruby
# db/seeds.rb

cable_ready.skip_updates do
  Chapter.destroy_all
  Chapter.create!
end
```

This would, of course, be way more flexible. On the other hand, it would require more thinking on the DX side, and thinking is hard.

I’d love to get your thoughts on this! 🤸‍♀️

leastbad:
I like these ideas, but I admit that I'm similarly stumped in terms of their implementation.

Is this something we could do on an environment basis? When are you running seeds? eg. is this a testing thing?

[...]

PR #168, opened 2021-12-02
Add `skip_updates` option

assuntaw:
This adds a `skip_updates` option to wrap any database CRUD changes in that should not be broadcast, for example because they are [happening in an environment without a running Redis server](https://github.com/stimulusreflex/cable_ready/issues/167).

```ruby
# app/models/application_record.rb
class ApplicationRecord < ActiveRecord::Base
  include CableReady::Updatable

  self.abstract_class = true
end

# db/seeds.rb

# Skips broadcasts for all models inheriting from ApplicationRecord:
ApplicationRecord.skip_updates do
  User.destroy_all
  User.create!(user_attrs)
end

# Skips broadcasts for User only:
User.skip_updates do
  # will not broadcast
  User.destroy_all
  User.create!(user_attrs)

  # will broadcast
  Post.destroy_all
  Post.create!(post_attrs)
end
```

Big thanks to @leastbad and @julianrubisch for pointing me in the right direction 🤗 I decided to borrow heavily from ActiveRecord's [no_touching](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/activerecord/lib/active_record/no_touching.rb#L23). Limiting `skip_updates` to the current thread (rather than working with a class variable) means that we don't skip broadcasts for other models or, crucially, other clients. I think that this is the safest option, and can't really think of a use case for one client prohibiting broadcasts for all other clients.

### DIFF
--- a/app/models/concerns/cable_ready/updatable.rb
+++ b/app/models/concerns/cable_ready/updatable.rb
@@ -25,6 +25,13 @@ module CableReady
           after_commit(ModelUpdatableCallbacks.new(:update, enabled_operations), {on: :update, if: options[:if]})
           after_commit(ModelUpdatableCallbacks.new(:destroy, enabled_operations), {on: :destroy, if: options[:if]})
         end
+
+        def self.skip_updates
+          skip_updates_classes.push(self)
+          yield
+        ensure
+          skip_updates_classes.pop
+        end
       end
     end
 
@@ -45,7 +52,7 @@ module CableReady
 
       def cable_ready_update_collection(resource, name, model)
         identifier = resource.to_global_id.to_s + ":" + name.to_s
-        ActionCable.server.broadcast(identifier, model.respond_to?(:previous_changes) ? {changed: model.previous_changes.keys} : {})
+        broadcast_updates(identifier, model.respond_to?(:previous_changes) ? {changed: model.previous_changes.keys} : {})
       end
 
       def enrich_association_with_updates(name, option)
@@ -92,6 +99,15 @@ module CableReady
           through_association: through_association,
           options: options
         })
+      end
+
+      def broadcast_updates(model_class, options)
+        return if skip_updates_classes.any? { |klass| klass >= self }
+        ActionCable.server.broadcast(model_class, options)
+      end
+
+      def skip_updates_classes
+        Thread.current[:skip_updates_classes] ||= []
       end
     end
   end
--- a/app/models/concerns/cable_ready/updatable/model_updatable_callbacks.rb
+++ b/app/models/concerns/cable_ready/updatable/model_updatable_callbacks.rb
@@ -15,13 +15,13 @@ module CableReady
       private
 
       def broadcast_create(model)
-        ActionCable.server.broadcast(model.class, {})
+        model.class.send(:broadcast_updates, model.class, {})
       end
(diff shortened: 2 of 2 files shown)


julianrubisch (approved):
LGTM!

app/models/concerns/cable_ready/updatable.rb
```diff
@@ -93,6 +100,15 @@ def enrich_association_with_updates(name, option)
           options: options
         })
       end
+
+      def broadcast_updates(model_class, options)
+        return if skip_updates_classes.any? { |klass| klass >= self }
```
leastbad, line 105: This is a really interesting line. I am happy to admit that I actually do not understand it. Can you break it down for me?

app/models/concerns/cable_ready/updatable.rb
```diff
@@ -93,6 +100,15 @@ def enrich_association_with_updates(name, option)
...
+        ActionCable.server.broadcast(model_class, options)
+      end
+
+      def skip_updates_classes
+        Thread.current[:skip_updates_classes] ||= []
```
leastbad, line 110: Elsewhere in the library, we use `Thread::Local`. Is that something we should look at, here?

https://github.com/socketry/thread-local/tree/master/guides/getting-started

leastbad (commented):
I apologize for the delay in getting this approved. I am very impressed with the simplicity of this solution. I would like to understand how it works a bit better so that I'll be able to support and potentially extend it in the future. However, this review is not intended as a blocker and I would be comfortable merging it right away if @julianrubisch is comfortable with everything. My understanding can come as a step 2 in this case. 😸

app/models/concerns/cable_ready/updatable.rb
julianrubisch: My understanding is that this checks for a thread local variable, whereas `Thread::Local` constructs a thread local singleton. IMHO these are two different things, but maybe @hopsoft can chime in?

leastbad: That tracks! I'm satisfied.

app/models/concerns/cable_ready/updatable.rb
assuntaw: Good question! I stole this directly from no_touching[https://github.com/rails/rails/blob/984c3ef2775781d47efa9f541ce570daa2434a80/activerecord/lib/active_record/no_touching.rb#L37](url). 

Essentially, this line makes sure that we do not broadcast if our `skip_updates_classes` include the current class or any class that the current class inherits from. Consider the following structure:

```ruby
class Book < ApplicationRecord
end

class Novel < Book
end
```

If we do the following:

```ruby
Book.skip_updates do
  Novel.first.update!
end
```

the update on `Novel.first` will not be broadcast because `Novel` inherits from `Book` and we decided to skip updates on `Book`.

To be honest, I wouldn't have come up with this if I hadn't read it in the `ActiveRecord` source code, but I do think it makes sense conceptually! 

What do you think?

leastbad: Well, TIL that `any?` supports a block argument, effectively turning it into a `select` that returns a Boolean. Cool!

And I know that you didn't technically come up with it, but I still assign the clever points to you. Thanks for the great explanation. I think that this is ready to go!

PR #169, opened 2021-12-13
fix: resolve error for webpack 5

acarpe:
the error is:
Should not import the named export 'version' (imported as 'version') from
default-exporting module (only default export is available soon)

### DIFF
--- a/javascript/index.js
+++ b/javascript/index.js
@@ -1,4 +1,4 @@
-import { version } from '../package.json'
+import packageInfo from '../package.json'
 import * as MorphCallbacks from './morph_callbacks'
 import { shouldMorphCallbacks, didMorphCallbacks } from './morph_callbacks'
 import * as Utils from './utils'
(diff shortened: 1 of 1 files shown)


leastbad:
Hello Antonio!

We have been discussing your PR and while it looks okay, it also seems odd that you're having any issue at all. We're actually going to be moving to a bundler system soon, too. I thought that in advance of changing the codebase, it makes sense to first understand if perhaps we can understand why the current setup is a problem for you. It's possible that we could fix it together in your project.

To paraphrase @marcoroth, ES modules are new-ish, but not so new that Webpack 5 shouldn't support them.

Are you running this with Webpacker 6 beta or something? I can say, for what it's worth, that Webpacker 5.4.3 works perfectly.

acarpe:
Yes I upgrade my rails app to use webpacker  6.0.0-rc.6

PR #170, opened 2021-12-18
Extract processBlock method and query getter

julianrubisch:
Houekeeping 🧹

### DIFF
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -41,9 +41,8 @@ export default class UpdatesForElement extends SubscribingElement {
   }
 
   async update (data) {
-    const identifier = this.getAttribute('identifier')
-    const query = `updates-for[identifier="${identifier}"]`
-    const blocks = document.querySelectorAll(query)
+    activeElement.set(document.activeElement)
+    const blocks = document.querySelectorAll(this.query)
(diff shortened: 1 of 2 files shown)


javascript/elements/updates_for_element.js
```diff
@@ -134,6 +134,10 @@ export default class UpdatesForElement extends SubscribingElement {
     )
   }
 
+  get query () {
+    return `updates-for[identifier="${this.getAttribute('identifier')}"]`
```
marcoroth, line 138: Can't you use `this.identifier` here too?
```suggestion
    return `updates-for[identifier="${this.identifier}"]`
```

javascript/elements/updates_for_element.js
```diff
@@ -41,9 +41,8 @@ export default class UpdatesForElement extends SubscribingElement {
   }
 
   async update (data) {
```
marcoroth, line 43: I think you can remove the `async` from the `update()` function now that you moved the async part to the `processBlock()` function

julianrubisch: good spot!

javascript/elements/updates_for_element.js
julianrubisch: 🙏

PR #171, opened 2021-12-23
Update .npmignore

Matt-Yorkley:
A bit of housekeeping; these files should not be present in the `npm` package.


marcoroth (approved):
Thank you @Matt-Yorkley!

PR #174, opened 2022-01-04
Ignore inner updates on a stimulus-reflex action or turbo submit

julianrubisch:
Let me preface this PR with the fact that this is already being used successfully in production and is intended as a backmerge of the work I'm doing at https://github.com/code-and-co/.

The idea is to opt out of an `updates_for` operation for the current user on a per-element basis, when a reflex or turbo submit is happening (for example, when wrapping a `<turbo-frame>`).

Conceptually, it consists of 3 parts:

### javascript/updatable/inner_updates_compat.js
which recursively marks every enclosing `updates-for` element with `performing-inner-update`, starting from a reflex or turbo submit initiating element. This attribute is removed again after completion.

### An `ignore_inner_update` option added to the helper
which adds a `ignore-inner-update` attribute to the element on demand.

### A shouldUpdate mechanism 
combining the existing "only" option and returning early when both `ignore-inner-updates` and `performing-inner-update` are set on an element.

cc @andrewerlanger @assuntaw

### DIFF
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -41,18 +41,9 @@ export default class UpdatesForElement extends SubscribingElement {
   }
 
   shouldUpdate (data) {
-    const only = this.getAttribute('only')
-
     return (
-      !(
-        this.hasAttribute('ignore-inner-updates') &&
-        this.hasAttribute('performing-inner-update')
-      ) &&
-      !(
-        only &&
-        data.changed &&
-        !only.split(' ').some(attribute => data.changed.includes(attribute))
-      ) &&
+      !this.ignoringInnerUpdates &&
+      this.hasChangesSelectedForUpdate(data) &&
       this.blocks[0] === this
     )
   }
@@ -145,6 +136,16 @@ export default class UpdatesForElement extends SubscribingElement {
     )
   }
 
+  hasChangesSelectedForUpdate (data) {
+    const only = this.getAttribute('only')
+
+    return !(
+      only &&
+      data.changed &&
+      !only.split(' ').some(attribute => data.changed.includes(attribute))
+    )
+  }
+
   get query () {
     return `updates-for[identifier="${this.identifier}"]`
   }
@@ -157,5 +158,12 @@ export default class UpdatesForElement extends SubscribingElement {
     return this.hasAttribute('debounce')
       ? parseInt(this.getAttribute('debounce'))
       : 20
+  }
+
+  get ignoringInnerUpdates () {
+    return (
+      this.hasAttribute('ignore-inner-updates') &&
+      this.hasAttribute('performing-inner-update')
+    )
   }
 }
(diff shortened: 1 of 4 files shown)


javascript/elements/updates_for_element.js
```diff
@@ -40,22 +40,33 @@ export default class UpdatesForElement extends SubscribingElement {
     }
   }
 
+  shouldUpdate (data) {
```
leastbad, line 43: lol - this function is so gnarly.

javascript/updatable/inner_updates_compat.js
```diff
@@ -0,0 +1,37 @@
...
+    recursiveMarkUpdatesForElements(event.detail.element)
+  })
+
+  document.addEventListener('stimulus-reflex:after', event => {
+    setTimeout(() => {
+      recursiveUnmarkUpdatesForElements(event.detail.element)
+    })
+  })
+
+  document.addEventListener('turbo:submit-start', event => {
```
leastbad, line 12: Should we consider adding event listeners for `turbolinks`?

leastbad (commented):
I sincerely hope that this is the last hairy edge case, for our sanity.

Anyhow: LGTM, moderated to a really liberal definition of "good" - I wish it wasn't necessary, but I agree with you that it is, and very much appreciate that you have championed getting this done properly.

javascript/updatable/inner_updates_compat.js
julianrubisch: My memory may not serve me well today but IIRC turbolinks had no submit-start and submit-end events

javascript/elements/updates_for_element.js
julianrubisch: Yeah. Actually now that you say it this could be more self-documenting, i.e. split into 3 getters...

javascript/updatable/inner_updates_compat.js
leastbad: Here's the list, FWIW: https://github.com/turbolinks/turbolinks#full-list-of-events

There's no `submit-start` or `submit-end`, but it's not clear if it's just been renamed or is an entirely new concept.

javascript/elements/updates_for_element.js
leastbad: Your call. I was just impressed at how it is like a homage to Perl.

javascript/updatable/inner_updates_compat.js
julianrubisch: Well since form handling via turbo is all-new I think probably the latter is the case.

And it really only applies to form submits anyway.

But you raise an interesting point about extension. What if we provided a way to add additional events to this list of listeners? 🤨

julianrubisch: Agreed

javascript/elements/updates_for_element.js
julianrubisch: updated

marcoroth (approved):

PR #175, opened 2022-01-04
Updated gemspec to specific Rails gems

paul:
Most apps don't require the full suite of Rails gems, and can avoid installing them to save CI and deploy time. This gem, however, had a dependency on the Rails meta-gem which ended up installing them all anyways.

Instead, we can limit it to only the Rails gems that we directly depend upon.

I'm not sure how to really test this, since the dummy app uses all of Rails. I grep'd through `./lib` to see what parts of Rails were being used, but that might not be comprehensive.

_Note_: I ran `./bin/standardize` as instructed by the README, but it touched dozens of files which I didn't commit as part of this PR.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,11 @@ PATH
   remote: .
   specs:
     cable_ready (5.0.0.pre8)
-      rails (>= 5.2)
+      actioncable (>= 5.2)
+      actionpack (>= 5.2)
+      activerecord (>= 5.2)
+      activesupport (>= 5.2)
+      railties (>= 5.2)
       thread-local (>= 1.1.0)
 
 GEM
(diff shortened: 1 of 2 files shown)


leastbad:
I added `ActionView` to the list, because we use it in `identifiable.rb`, as well as the test suite.

The bigger question which @julianrubisch is better-equipped to answer is whether the test dummy requiring all of Rails means that the Gemspec for CableReady needs to include all of Rails.

leastbad:
Actually, I hope I haven't made a mess of this - I just realized that we'll have to run `bundle` on this so that the `Gemfile.lock` is regenerated. Can you please pull it, run `bundle` and push the changes? With my apologies and graditude.

julianrubisch:
> The bigger question which @julianrubisch is better-equipped to answer is whether the test dummy requiring all of Rails means that the Gemspec for CableReady needs to include all of Rails.

Well the easiest way to tell is to, well, run the tests. But either way I think it could be added as a development dependency...

leastbad:
> Well the easiest way to tell is to, well, run the tests. But either way I think it could be added as a development dependency...

LGTM:
```
Finished in 8.717977s, 6.4235 runs/s, 19.2705 assertions/s.
56 runs, 168 assertions, 0 failures, 0 errors, 0 skips
```

paul:
I did add it as a development dependency: https://github.com/stimulusreflex/cable_ready/pull/175/files#diff-2f4da7b98a9c86d38f368f6119614e27dbb9c1e4dc3d70240cb466cc211435acR32

PR #176, opened 2022-01-06
Fix `text_content` operation for nullish values

marcoroth:
## Type of PR
Bug Fix

### Description

This PR adds a check to make sure that we use the provided `text` parameter if it's not nullish

### Why should this be added

The `text_content` operation always used an empty string for the `text` parameter if the value of the `text` parameter was nullish.

Previously if you passed in `0` as a value for `text` it would have still used the empty string instead.

### Checklist

### DIFF
--- a/javascript/operations.js
+++ b/javascript/operations.js
@@ -159,7 +159,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { text, focusSelector } = operation
-        element.textContent = (text != null) ? text || ''
+        element.textContent = (text != null) ? text : ''
         assignFocus(focusSelector)
       })
       after(element, operation)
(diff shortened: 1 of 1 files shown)


javascript/operations.js
```diff
@@ -159,7 +159,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { text, focusSelector } = operation
-        element.textContent = text || ''
+        element.textContent = (text != null) ? text || ''
```
ParamagicDev, line 162:
```suggestion
        element.textContent = (text != null) ? text : ''
```

marcoroth: 🤦🏼

PR #179, opened 2022-01-31
CableReady installer

julianrubisch:
I've taken a first stab at the _new unified CableReady Installer_, and borrowed heavily from https://github.com/hotwired/turbo-rails/blob/main/lib/tasks/turbo_tasks.rake

Let me break it down for you a bit:

- basically we have 2 install tasks, `cable_ready:install:importmap`, and `cable_ready:install:node`, where the main `cable_ready:install` task switches on the presence of an `importmap.rb` file
- both installers do a sanity check if Redis and ActionCable are available, and optionally asks the user if Redis should be installed (https://github.com/stimulusreflex/cable_ready/pull/179/files#diff-e469da9bc01523658da7d880f0c454118061110611e92e21c3d69e0232ae983aR8)
- the installers then do their `pin`/`yarn add` actions
- finally, initializing CableReady in the application's entrypoint. I've taken my own experience as a heuristic here:
  1. `import CableReady from "cable_ready"
  2. if a `./channels/index.js` is present, we're likely in a webpacker setup. Use that consumer to initialize CR.
  3. if no such channels index is present, we can fall back on using the `cable` module exposed by `@hotwired/turbo-rails`.
  4. else ask the user to do it manually.

I've tested many scenarios locally, but doing a bit of testing certainly wouldn't hurt 🙏

### DIFF
--- a/lib/install/cable_ready_entrypoint.rb
+++ b/lib/install/cable_ready_entrypoint.rb
@@ -1,0 +1,36 @@
+if (js_entrypoint_path = Rails.root.join("app/javascript/application.js")).exist?
+  entrypoint_content = File.read(js_entrypoint_path)
+
+  say "Import CableReady"
+  append_to_file js_entrypoint_path, %(import CableReady from "cable_ready"\n)
+
+  say "Initializing CableReady with ActionCable consumer"
+
+  # Check if there's a channels file, i.e. we're in a webpacker or esbuild setup
+  if Rails.root.join("app/javascript/channels/index.js").exist?
+    unless entrypoint_content.match?(/import consumer/)
+      append_to_file js_entrypoint_path, %(import consumer from "./channels/consumer"\n)
+    end
+
+    append_to_file js_entrypoint_path, %(CableReady.initialize({ consumer })\n)
+
+  # else we could use the cable.consumer from turbo-rails, if present
+  elsif entrypoint_content.match?(/import.*@hotwired\/turbo-rails/)
+    unless entrypoint_content.match?(/import.*cable.*@hotwired\/turbo-rails/)
+      append_to_file js_entrypoint_path, %(import { cable } from "@hotwired\/turbo-rails"\n)
+    end
+
+    append_to_file js_entrypoint_path, <<~JS
+      (async () => {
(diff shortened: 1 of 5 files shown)


Matt-Yorkley:
Hey :wave:

I've been playing with CR and SR on a greenfield app that's using Vite and ran into a couple of issues with some of the install tasks for both. I took some rough notes:

The helpers_generator here fails with Vite:
https://github.com/stimulusreflex/cable_ready/blob/1bded053b64e8f497fcf368a258b0326ea4c886f/lib/generators/cable_ready/helpers_generator.rb#L12-L22
It assumes the `main_path` is "app/javascript" and the entrypoint directory is named "packs". 

A slight adjustment makes it work for either:
```diff
    main_folder = defined?(Webpacker) ? Webpacker.config.source_path.to_s.gsub("#{Rails.root}/", "") : "app/javascript"
++  main_folder = "app/frontend" if defined? ViteRuby

    filepath = [
      "#{main_folder}/controllers/index.js",
      "#{main_folder}/controllers/index.ts",
      "#{main_folder}/packs/application.js",
      "#{main_folder}/packs/application.ts",
++    "#{main_folder}/entrypoints/application.js",
++    "#{main_folder}/entrypoints/application.ts"
    ]
```
There are some similar install tasks in SR that also break in Vite but just need a very slightly different path. And there's this Webpacker-specific line that gets added during setup:
```js
StimulusReflex.debug = process.env.RAILS_ENV === 'development'
```
which throws an error with Vite. It works with a slight (Vite-specific) adjustment:
```diff
--StimulusReflex.debug = process.env.RAILS_ENV === 'development'
++StimulusReflex.debug = import.meta.env.DEV
```

Matt-Yorkley:
Obviously it's annoying AF having to cater to Webpacker / Shakapacker(!) / Importmaps / jsbundling-rails / Vite, but from the DX side of things it's really nice if the install tasks Just Work™ without any issues :man_shrugging:

julianrubisch:
Yeah, so this PR is about the main CR installer (which even doesn’t exist yet!)

shall we pull out your comment into an issue?

julianrubisch:
Closing in favor of #233

PR #182, opened 2022-02-13
updates-for should respect url and only params on each element

leastbad:
Reworked updates-for element to allow per-block url and only attributes. URL fragment indexes are now tracked on a per-URL basis.

That means the following contrived example works as you'd expect:

`index.html.erb`
```erb
<%= updates_for current_user, :posts do %>
  <% Post.all.each do |p| %>
    <p><%= p.name %>, <%= p.email %></p>
  <% end %>
  <p><%= rand(1..1000) %></p>
<% end %>

<%= updates_for current_user, :posts, only: [:name] do %>
  <% Post.all.each do |p| %>
    <p><%= p.name %>, <%= p.email %></p>
  <% end %>
  <p><%= rand(1..1000) %></p>
<% end %>

<%= updates_for current_user, :posts, url: "/updates_for" do %>
  <% Post.all.each do |p| %>
    <p><%= p.name %>, <%= p.email %></p>
  <% end %>
  <p><%= rand(1..1000) %></p>
<% end %>

<%= updates_for current_user, :posts, only: [:name], url: "/updates_for" do %>
  <% Post.all.each do |p| %>
    <p><%= p.name %>, <%= p.email %></p>
  <% end %>
  <p><%= rand(1..1000) %></p>
<% end %>
```
`updates_for.html.erb`
```erb
<%= updates_for current_user, :posts do %>
  <p>HAIRY!</p>
  <p><%= rand(1..1000) %></p>
<% end %>
```
No, I don't know why `Post` has a `name` and `email`. Don't ask hard questions.
```ruby
post = current_user.posts.create name: "Fred", email: "fred@gmail.com"
post.update name: "Frederick" # updates the first three blocks, plus console warning
post.update email: "frederick@gmail.com" # updates only the first and third
```
Note that the 4th block will not be updated because `updates_for.html.erb` only has one block; a warning is generated in the console to suggest that there wasn't enough elements for the update to complete.

Unfortunately this will certainly clash with #181 as written, although I have done my best to maintain the general structure and naming to ease merge conflicts.

### DIFF
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -46,9 +46,6 @@ export default class UpdatesForElement extends SubscribingElement {
   }
 
   update (data) {
-    // memoize blocks to avoid unnecessary DOM traversal
-    this.blocks = document.querySelectorAll(this.query)
-
     // first updates-for element in the DOM *at any given moment* updates all of the others
     if (this.blocks[0] !== this) return
 
@@ -158,7 +155,7 @@ export default class UpdatesForElement extends SubscribingElement {
   shouldUpdate (data, block) {
     // if everything that could prevent an update is false, update this block
     return (
-      this.processInnerUpdates(block) &&
+      !this.ignoresInnerUpdates(block) &&
       this.hasChangesSelectedForUpdate(data, block)
     )
   }
@@ -174,12 +171,19 @@ export default class UpdatesForElement extends SubscribingElement {
     )
   }
 
-  processInnerUpdates (block) {
+  ignoresInnerUpdates (block) {
     // don't update during a Reflex or Turbolinks redraw
-    return !(
+    return (
       block.hasAttribute('ignore-inner-updates') &&
       block.hasAttribute('performing-inner-update')
     )
+  }
+
+  get blocks () {
+    // memoize blocks to avoid unnecessary DOM traversal
+    if (!this._blocks) this._blocks = document.querySelectorAll(this.query)
+
+    return this._blocks
   }
 
   get query () {
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -45,45 +45,58 @@ export default class UpdatesForElement extends SubscribingElement {
     }
   }
 
-  shouldUpdate (data) {
-    return (
-      !this.ignoringInnerUpdates &&
-      this.hasChangesSelectedForUpdate(data) &&
-      this.blocks[0] === this
-    )
-  }
-
   update (data) {
+    // memoize blocks to avoid unnecessary DOM traversal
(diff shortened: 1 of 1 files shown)


javascript/elements/updates_for_element.js
```diff
@@ -45,45 +45,58 @@ export default class UpdatesForElement extends SubscribingElement {
...
-      !this.ignoringInnerUpdates &&
-      this.hasChangesSelectedForUpdate(data) &&
-      this.blocks[0] === this
-    )
-  }
-
   update (data) {
-    ActiveElement.set(document.activeElement)
+    // memoize blocks to avoid unnecessary DOM traversal
+    this.blocks = document.querySelectorAll(this.query)
```
julianrubisch, line 50: I'd still move this to a `get`ter for better structure, but memoized, i.e.

```js
get blocks() {
  if(!this._blocks) this._blocks = document.querySelectorAll(this.query)
  
  return this._blocks
}
```

javascript/elements/updates_for_element.js
```diff
@@ -152,24 +174,21 @@ export default class UpdatesForElement extends SubscribingElement {
     )
   }
 
-  get query () {
-    return `updates-for[identifier="${this.identifier}"]`
+  processInnerUpdates (block) {
```
julianrubisch, line 177: I'd probably rename this to something that better indicates it's a predicate (right now it sounds like a command), e.g. `processesInnerUpdates` or other way round `ignoresInnerUpdates`. The latter would be better aligned with the attribute `ignore-inner-updates`, fwiw.

julianrubisch (commented):
A first set of comments... Also, I'd love to know why the dev builds job failed, because I'd love to test this against an app I have here... @marcoroth any idea ?

javascript/elements/updates_for_element.js
leastbad: My hope is to hand this off again, soon - and if you want it to be a getter, it shall be a getter! I localized it for the development because it's only called in twice, and in the same method, too.

javascript/elements/updates_for_element.js
leastbad: Sounds good to me. I changed it to simplify the logic (every negative makes comprehension one step harder) but I actually wasn't even aware that there was an `ignore-inner-updates` attribute. Makes good sense, indeed, now that I know.

javascript/elements/updates_for_element.js
julianrubisch: I just love short methods 😅

leastbad:
In the interim, I've had good luck running `yarn run watch` and then pointing my `packages.json` dependencies at `"link:/home/leastbad/cable_ready/dist/cable_ready.umd.js"`.

PR #184, opened 2022-02-14
Add debug log output

julianrubisch:
I just added debug output and now it... works?

### DIFF
--- a/.github/workflows/dev-build.yml
+++ b/.github/workflows/dev-build.yml
@@ -32,8 +32,12 @@ jobs:
           
           tar -xf cable_ready-dev-build.tgz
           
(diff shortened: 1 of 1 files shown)


marcoroth:
This one worked because the source branch of this PR is from this repo. The other one didn't work because that PR had the source branch from a fork of this repo. 

GitHub doesn't share the secrets across forks for security reasons. That's also why the user/credentials were missing and thus resulting in the failed action run.

See: https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow 
> Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

PR #187, opened 2022-02-16
Export utils

julianrubisch:
Export `kebabize` so it can be reused in child libraries (SR, futurism, cubism etc.) thus saving on bundle size.

### DIFF
--- a/javascript/utils.js
+++ b/javascript/utils.js
@@ -149,5 +149,7 @@ export {
   after,
   debounce,
   handleErrors,
-  graciouslyFetch
+  graciouslyFetch,
+  debounce,
+  kebabize
 }


javascript/utils.js
```diff
@@ -149,5 +149,7 @@ export {
   after,
   debounce,
   handleErrors,
-  graciouslyFetch
+  graciouslyFetch,
+  debounce,
```
marcoroth, line 153: `debounce` is already being exported on L150

```suggestion
```

marcoroth (changes requested):

javascript/utils.js
julianrubisch: Yeah I figured, that's why I... oh, forget it 🤦

PR #188, opened 2022-02-18
Extract Block class and do a pre-fetch pass

julianrubisch:
Mainly this pulls together everything that reaches out to `block` in its own class, and, more importantly, does a pre-pass of fetching unique URLs

While this is an improvement in general and might fix some of the multi-fetches occurring, I still saw some instances where the same URL was fetched twice, ~even from the same elements~.

What I can say though is that in the app I'm testing it against, we have a fair amount of `touch`es happening in the class hierarchy, so I think what we're _really_ looking at is another opportunity to look at **server side debouncing**

cc @andrewerlanger

### DIFF
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -45,9 +45,9 @@ export default class UpdatesForElement extends SubscribingElement {
     }
   }
 
-  update (data) {
+  async update (data) {
     // first updates-for element in the DOM *at any given moment* updates all of the others
-    if (this.blocks[0] !== this) return
+    if (this.blocks[0].element !== this) return
 
     // hold a reference to the active element so that it can be restored after the morph
     ActiveElement.set(document.activeElement)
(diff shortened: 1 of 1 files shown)


javascript/elements/updates_for_element.js
```diff
@@ -45,71 +45,105 @@ export default class UpdatesForElement extends SubscribingElement {
...
+      block.process(data, this.html, this.index)
+    })
   }
 
-  async processBlock (data, block) {
-    // memoize the block's URL to avoid unnecessary DOM interop
-    const blockURL = url(block)
+  get blocks () {
+    // memoize blocks to avoid unnecessary DOM traversal
+    if (!this._blocks)
```
leastbad, line 86: Will memoizing `blocks` cause issues if the contents of the page changes? For example, if a second block was morphed into the page without disturbing this element, it seems like the 2nd block would never get operated upon.

In version I made, before I pulled `blocks` out into a getter, I made a local copy of the current `blocks` snapshot inside of the `update` method. It was only intended to exist for the duration of the `update` call.

Now that we have an asynch `update`, we need to be really careful about what we're memoizing. I actually do think we need to revert this to something that exists only during the life of one `update` call.

javascript/elements/updates_for_element.js
```diff
@@ -45,71 +45,105 @@ export default class UpdatesForElement extends SubscribingElement {
...
+    template.innerHTML = String(html[this.url]).trim()
 
     await this.resolveTurboFrames(template.content)
 
     const fragments = template.content.querySelectorAll(this.query)
 
-    if (fragments.length <= index) {
-      console.warn('Update aborted due to insufficient updates-for elements')
+    if (fragments.length <= blockIndex) {
+      console.warn('Update aborted due to mismatched number of elements')
```
leastbad, line 130: I am fine with changing the language, although please consider that this will only catch scenarios where there are not enough blocks to complete the update present. If there are more than expected, they will just be ignored.

I wonder if it makes sense to include the block url in the warning message? It might assist the developer in tracking down which page is problematic.

leastbad (changes requested):
I think that we're **really** close. Great work so far.

javascript/elements/updates_for_element.js
julianrubisch: This was just refactored by me... I'm fine with de-memoizing it, but I think that should be it, then it's just a simple wrapper around `document.querySelectorAll`

julianrubisch: okay, I think now get you... let me just quickly fix this

leastbad:
Thanks for getting this pushed through!

PR #193, opened 2022-03-26
Include changeset in class-level updates broadcast

### LINKED ISSUE #192: Inlcude changed keys when broadcasting to model class? (updates_for :only)
My use case is that I would like to put this in a view:

```erb
<%= updates_for StaffMember, only: 'current_task' do %>
 <%# show what people are working on #%>
<% end %>
```

That is: trigger a refresh when the `current_task` of any StaffMember changes. However, it seems like broadcasts to the subscription ID of a model class are always an empty hash, so the `:only` option is effectively ignored.

I'm pretty sure that's happening here: https://github.com/stimulusreflex/cable_ready/blob/40bc658cc500b0715dfb9fb49cf539f23f24f98c/app/models/concerns/cable_ready/updatable/model_updatable_callbacks.rb#L22-L25

Could we include the changed attributes when broadcasting to the `model.class` as well? I feel like it would at least be harmless, but maybe I've missed the reasoning for doing it this way.

grncdr:
Enables using the :only option with updates_for to filter updates when subscribing to all instances of a model class.

Fixes #192


julianrubisch:
Okay I expected that test failure, that's okay.

There also don't seem to be any tests for this behavior yet which is an omission on our part. I'll write some tests for this soonish.

The feature itself looks sensible, though I'd also like to ask @andrewerlanger for his opinion since he originally proposed this feature and maybe has a fresher memory of things than me.

grncdr:
@julianrubisch I'm not sure what you mean by "There also don't seem to be any tests for this behavior yet". I looked at the test failures, and they were mostly/all tests for this behavior, so I've updated the expectations.

julianrubisch:
Sorry for being unclear - I meant there's currently no tests for a specific case testing the `only` option. Not related to this PR at all.

I'd still love for @andrewerlanger to check back on this in a spare minute to make sure I don't miss something!

andrewerlanger:
@grncdr thanks very much for taking this on 🙏 

The code itself looks good to me, I'm just not sure whether the following is true:

> when subscribing to all instances of a model class.

As far as I understand: when passing a class constant to `updates_for`, you don't fully subscribe to all instances of that class, but rather to `create` and `destroy` callbacks specifically. This is what makes it a good candidate for `index` views.

If that is indeed the case, it would mean that the changes introduced in this PR do the following for the code below:

```rb
<%= updates_for Item, only: :title do %>
  <% items.each do |item| %>
    <p><%= item.title %></p>
  <% end %>
<% end %>
```

- If an item is created with a `title`: it gets appended to the list
- If an item is created without a `title`: the list remains unchanged
- If you update an item's `title`: the list remains unchanged

From the PR description, it sounds to me like that last point is the important one. If you update an item's `title`, the list _should_ be changed (specifically, the new title should be displayed). But I don't think that will be the case here.

I could very well be wrong about this though, so all the more reason to add some tests :)

grncdr:
Thanks for checking this out @andrewerlanger  👍 

> From the PR description, it sounds to me like that last point is the important one. If you update an item's title, the list should be changed (specifically, the new title should be displayed). But I don't think that will be the case here.

There's some confusion here about what problem this PR is actually solving, probably because I didn't include the issue description from #192 here. My issue is not that updates were being *missed*, but that updates_for was causing (many) reloads for updates that were _uninteresting_. Sticking with the `item.title` example, imagine if `Item` also contained a `view_counter` that was being updated every few seconds. Given your example code, everybody viewing items is going to reload the list every few seconds, even though the item titles have never changed.

grncdr:
What follows is a lot of pedantic clarification, feel free to skip it if we're already on the same page.

> As far as I understand: when passing a class constant to updates_for, you don't fully subscribe to all instances of that class, but rather to create and destroy callbacks specifically.

That is not the case. If the model class contains `enable_updates` and the `:on` option includes `::update` (which it will by default), it will broadcast to the `model.class` channel after any update on the first line of `broadcast_update`: https://github.com/stimulusreflex/cable_ready/blob/50ffe496eefdbd041dec3dfc99c97c240c41ad12/app/models/concerns/cable_ready/updatable/model_updatable_callbacks.rb#L22-L25 (this commit is the main branch at the time I made the comment).

The `model.class` channel is of course the same channel that `updates_for Item do ... end` is subscribed to, but because the broadcast data does not contain the "changed" key, the `only:` option is ignored: https://github.com/stimulusreflex/cable_ready/blob/50ffe496eefdbd041dec3dfc99c97c240c41ad12/javascript/elements/updates_for_element.js#L190-L199

Regarding the scenarios you described:

> * If an item is created with a `title`: it gets appended to the list 
> * If an item is created without a `title`: the list remains unchanged

The broadcasting of updates for items being created (or destroyed) goes through a different code path, calling [`broadcast_create`](https://github.com/stimulusreflex/cable_ready/blob/50ffe496eefdbd041dec3dfc99c97c240c41ad12/app/models/concerns/cable_ready/updatable/model_updatable_callbacks.rb#L17) unless the user has specifically opted out with the `on:` option to `enable_updates`.

> * If you update an item's `title`: the list remains unchanged

This is also incorrect, the list will be refreshed, both in the current version and in the version including my changes. However, if some _other_ attribute of the Item is updated, the list will be refreshed in the current version, and *won't* be refreshed after my PR.

All this PR does is add a little more information to that existing broadcast so that the `updates_for` component can ignore updates it doesn't care about.

andrewerlanger:
@grncdr thanks for clarifying, it sounds like I was wrong in that case. I thought something like `<% updates_for Item %>` would only hook into the `create` and `destroy` callbacks, which would have meant an update to e.g. `item.view_counter` wouldn't interfere with the list of items.

If, as you suggest, all items are indeed being updated for _uninteresting_ updates, then I fully agree it would be useful to opt out of these. That was also my motivation for proposing an `only:` option at the level of individual model instances in the first place.

Appreciate you taking the time to explain this in detail. Everything looks good to me with the help of that context (although with the minor caveat that I haven't had a chance to test it myself yet). 

@julianrubisch @leastbad what do you think?

julianrubisch:
Thanks @grncdr for the detailed rundown. 💚

I admit that even though writing most of it I'm sometimes fuzzy wrt the code paths at play.

That part is also hard to test because it involves an interaction of backend and frontend code.

I will reserve some time soon to write a couple of JavaScript tests that should hopefully clear things up for the next time I need to look at it 😬

PR #198, opened 2022-04-18
Add an `invoke_method` operation

jaredcwhite:
This adds an operation which allows you to call a JavaScript method for an element, whether that's a builtin HTML element or a custom element. Or you can use `window` as the receiver instead of an element. Example:

```ruby
cable_car.invoke_method "#some-element", method: :doSomething,
                                         args: ["wow", { very: "cool" }], # you can (optionally) pass any number of args via an array
                                         delay: 1000 # as with all operations now, you can delay it by X ms

cable_car.invoke_method receiver: :window, method: "Turbo.visit", args: ["/path"]
```

- [ ] Documentation
- [ ] Tests (not sure how?)

### DIFF
--- a/javascript/operations.js
+++ b/javascript/operations.js
@@ -179,16 +179,23 @@ export default {
     })
   },
 
-  callMethod: operation => {
+  invokeMethod: operation => {
     processElements(operation, element => {
       before(element, operation)
       operate(operation, () => {
-        const { name, args } = operation
-        if (!element[name]) {
-          console.warn(`CableReady callMethod failed due to missing '${name}' method for element:`, element)
-          return
+        const { element, receiver, method, args } = operation
+        const chain = method.split('.')
+        const obj = receiver === 'window' ? window : element
+        const foundMethod = chain.reduce((lastTerm, nextTerm) => (lastTerm[nextTerm] || {}), obj)
+
+        if (foundMethod instanceof Function) {
+          foundMethod(...(args || []))
+        } else {
+          console.warn(
+            `CableReady invokeMethod failed due to missing '${method}' method for:`,
+            receiver === 'window' ? window : element
+          )
         }
-        element[name](...(args || []))
       })
       after(element, operation)
     })
--- a/javascript/operations.js
+++ b/javascript/operations.js
@@ -207,7 +207,7 @@ export default {
           foundMethod.apply(lastObjectInChain, args || [])
         } else {
           console.warn(
-            `CableReady invoke_method failed due to missing '${method}' method for:`, firstObjectInChain
+            `CableReady invoke_method operation failed due to missing '${method}' method for:`, firstObjectInChain
           )
         }
       })
--- a/javascript/operations.js
+++ b/javascript/operations.js
@@ -188,7 +188,7 @@ export default {
           console.warn(`CableReady callMethod failed due to missing '${name}' method for element:`, element)
           return
         }
-        element[name](...args)
+        element[name](...(args || []))
       })
       after(element, operation)
     })
--- a/javascript/operations.js
+++ b/javascript/operations.js
@@ -183,17 +183,31 @@ export default {
     processElements(operation, element => {
       before(element, operation)
       operate(operation, () => {
+        let firstObjectInChain
         const { element, receiver, method, args } = operation
         const chain = method.split('.')
-        const obj = receiver === 'window' ? window : element
-        const foundMethod = chain.reduce((lastTerm, nextTerm) => (lastTerm[nextTerm] || {}), obj)
-
+    
+        switch (receiver) {
+          case 'window':
+            firstObjectInChain = window
+            break;
+          case 'document':
+            firstObjectInChain = document
+            break;
+          default:
+            firstObjectInChain = element
+        }
+        let lastObjectInChain = firstObjectInChain
+        const foundMethod = chain.reduce((lastTerm, nextTerm) => {
+          lastObjectInChain = lastTerm
+          return lastTerm[nextTerm] || {}
+        }, firstObjectInChain)
+    
         if (foundMethod instanceof Function) {
-          foundMethod(...(args || []))
+          foundMethod.apply(lastObjectInChain, args || [])
         } else {
           console.warn(
-            `CableReady invokeMethod failed due to missing '${method}' method for:`,
-            receiver === 'window' ? window : element
+            `CableReady invoke_method failed due to missing '${method}' method for:`, firstObjectInChain
           )
         }
       })
--- a/javascript/operations.js
+++ b/javascript/operations.js
@@ -179,6 +179,21 @@ export default {
     })
   },
 
+  callMethod: operation => {
+    processElements(operation, element => {
+      before(element, operation)
+      operate(operation, () => {
+        const { name, args } = operation
+        if (!element[name]) {
+          console.warn(`CableReady callMethod failed due to missing '${name}' method for element:`, element)
+          return
+        }
+        element[name](...args)
+      })
+      after(element, operation)
+    })
+  },
+
   removeAttribute: operation => {
     processElements(operation, element => {
       before(element, operation)
--- a/lib/cable_ready/config.rb
+++ b/lib/cable_ready/config.rb
@@ -40,7 +40,6 @@ module CableReady
       Set.new(%i[
         add_css_class
         append
-        call_method
         clear_storage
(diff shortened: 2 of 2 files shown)


marcoroth:
I love this! ❤️ 

I'm wondering if we additionally should support a way of invoking a function on the `window` or `document` object instead of "just" elements.

leastbad:
> I'm wondering if we additionally should support a way of invoking a function on the `window` or `document` object instead of "just" elements.

100% but what's the least janky API? If there was only one, we could structure it as a boolean that falls through to the selector as default.

We could accept a string option called `target` that defaults to `selector` but could also be `window` or `document`, and then we use a `switch` that falls through to the selector?

leastbad:
... or a string option called `receiver`, which I suspect is more intuitive than `target` in this context.

leastbad:
That is a good point, but technically isn't this already an issue with selector-based objects as well?

A Web Component could expose an Object that has several methods, no?

Part of the reason I suggested `receiver` is because there's essentially three root objects that might hold function references. It seems like the question of which one might deserve to stand alone from the question of nested objects.

---

As it happens, I have a theoretical solution / syntax to propose. Note that I'm not lobbying for this approach, just stating my first instinct of how I might tackle it.

Once your `receiver` is established, then it seems like we could just split on the `.` character and construct a multi-level call:

```
cable_ready operation: "call_method", receiver: "window", method: "Foo.Bar.baz"
```

I think this is pretty reasonable, and no more exploitable than a script tag already is.

jaredcwhite:
@leastbad dayyuum, that's tempting. 😄

hopsoft:
I wonder if that last example would feel more appropriate as:

```rb
cable_ready operation: "call_method", receiver: "window.Foo.Bar", method: "baz"
```

leastbad:
I am just chewing over all of this.

We have three potential options in play, because we shant forget `selector`.

Actually, let me back up...

In CR, most operations operate on a `selector`, but a minority ignore `selector` and are hard-wired to operate on `window` or `document`. The important takeaway here is that **you cannot specify `window` or `document` as a `selector`**. There's no logic [at present?] to parse out `window` and `document` and handle them as JS objects, instead of DOM elements.

This is what lead to me proposing a `receiver`; otherwise, we could have just used the `selector`.

My goals here are to expose functionality in a maximally useful way that is consistent with the call signatures of the other operations. I believe we also share a powerful motivation to eliminate confusion without restricting creativity. I bring this up because there's a sharp edge that I am actively trying to round off:

```ruby
# bad, in my opinion
cable_ready operation: "call_method", selector: "#test", receiver: "Foo.Bar", method: "baz"
cable_ready operation: "call_method", receiver: "window.Foo.Bar", method: "baz"

# good, in my opinion
cable_ready operation: "call_method", selector: "#test", method: "Foo.Bar.baz"
cable_ready operation: "call_method", receiver: "window", method: "Foo.Bar.baz"
```

Do you see what I see? I am trying to avoid a "if you're doing a DOM element, no receiver is required but if you're working with window or document, you include the name of the object *as a string* inside of the receiver" scenario. That, to me, does not pass the sniff test.

jaredcwhite:
Maybe even better:

```ruby
cable_ready operation: "invoke", method: "Foo.Bar.baz"
```

@hopsoft and some others suggested `invoke` which is growing on me

and we don't need `receiver` if we assume that it's always `window` unless a selector is specified. even `document` is easily accessible as a member of `window`

jaredcwhite:
I'm also realizing this re-raises the specter of "is it a method or a function?" …because if you declare a global function (setting ESM/imports stuff aside):

```js
function printStr(str) {
  console.log(str)
}
```

it becomes a method on the `window` object. 😄

```js
window.printStr("Hello world")
```

which you could then invoke… 🧐

leastbad:
> @hopsoft and some others suggested `invoke` which is growing on me

This operation will be called `invoke_method`. @hopsoft has always historically been the decider in these moments, and with the exception of "nothing morph" (still makes my head spin :wink:) he's proven pretty damn wise in this regards. My bias is always towards shorter names, but a consistent, guessable API (see `dispatch_event`) has to take priority. I'm actually happy to have the naming locked in so that we can focus on the fun part, API-wise.

> we don't need receiver if we assume that it's always window unless a selector is specified

Oh, man... you had me super excited for about two minutes. The problem is that the way CR works - and it's been this way for a long time - is that an undeclared `selector` falls through to `document`. So there's no way to just assume `window` because we have no way to know if they *specified* `document`, or if it was just the default.

Still, I love where you were headed with this. Not having to specify a `receiver` would be pretty nice. Technically, right now you don't have to specify a `selector` OR a `receiver` when you're invoking methods on `document`. Too bad that people don't tend to put methods there. 🎭 

If you have any other ideas that would get us there, keep 'em coming.

jaredcwhite:
Ah, that document fallback totally escaped me. Well…I think I'm on board with your syntax then!

leastbad:
Do you want to take a crack at the implementation? Happy to help but don't want to step on toes.

jaredcwhite:
Yup

marcoroth:
Maybe I'm not following 100%, but if you have this call...
```ruby
cable_ready.invoke_method(receiver: "window", method: "Foo.Bar.baz")
```
... you are assuming that `Bar` and `baz` are properties on the respective object. 

But what if those are not actual properties, but functions? Functions which might even take arguments? Should we support something like this as well?

In JavaScript this could look like:
```js
Foo.Bar.getBaz().doSomethingOnBaz("with arguments", 2)
```

I feel like method chaining would be the "most natural" in that case. Maybe, something like:
```ruby
cable_ready
  .invoke_method(receiver: "window", method: "Foo.Bar.getBaz")
  .invoke_method(chain: true, method: "doSomethingOnBaz", args: ["with arguments", 2])
```

Is this too far?

leastbad:
Strong first reaction: that's *a lot*. I think that at some point, you need a sane cut-off before it makes more sense to write a custom operation or just call something that calls something else.

Is this partially motivated by having no idea how this would actually be implemented? Sure. But mostly it just seems complicated AF for a vanishingly small surface area of opportunity.

jaredcwhite:
Well, speaking only for myself, the number of times I expect to call methods off of global JS objects is roughly, er, never. 😄 My original use case was wanting to trigger a custom element to do something via a method call, so that doesn't even need any method chaining. I do like the idea of method chaining, but only in the simple property-based form until you get to the final method signature. I think beyond that it becomes too fiddly.

jaredcwhite:
I wonder if we should also update `setProperty` so it supports a window receiver as well as nested `foo.bar.baz` functionality.

javascript/operations.js
```diff
@@ -179,16 +179,23 @@ export default {
...
     processElements(operation, element => {
       before(element, operation)
       operate(operation, () => {
-        const { name, args } = operation
-        if (!element[name]) {
-          console.warn(`CableReady callMethod failed due to missing '${name}' method for element:`, element)
-          return
+        const { element, receiver, method, args } = operation
+        const chain = method.split('.')
+        const obj = receiver === 'window' ? window : element
```
leastbad, line 188: Sadly, I don't think we can get away with a ternary here since we have no way to specify `document`.

javascript/operations.js
```diff
@@ -179,16 +179,23 @@ export default {
...
+        const foundMethod = chain.reduce((lastTerm, nextTerm) => (lastTerm[nextTerm] || {}), obj)
+
+        if (foundMethod instanceof Function) {
+          foundMethod(...(args || []))
+        } else {
+          console.warn(
+            `CableReady invokeMethod failed due to missing '${method}' method for:`,
```
leastbad, line 195: Can we tweak this from "invokeMethod" to "invoke_method operation", please?

javascript/operations.js
```diff
@@ -179,16 +179,23 @@ export default {
...
+            receiver === 'window' ? window : element
```
leastbad, line 196: Same ternary issue as earlier. We have to support `document`, as there's no way to specify it with a selector.

leastbad (changes requested):
We're close! Great work on this one so far.

jaredcwhite:
Sounds good, I'll try to get to these changes soon. Been outta commission for a few days now, not sure exactly when I'll be back to full speed ahead…

leastbad:
This one is close, too. 😉

leastbad (approved):

PR #204, opened 2022-05-18
Register STI descendants for collection updates

julianrubisch:
## Problem

I noticed that when working with STI models in a `has_many` relation, the collection callbacks would only trigger updates for the base class, not its descendants. So in short:

```rb
class Section
  has_many :blocks, enable_updates: true
end

class Block
  belongs_to :section
end

class Comment < Block; end
```

```erb
<%= updates_for @section, :blocks do %><%= render @section.blocks %><% end %>
```

would only ever trigger updates if I added an instance of `Block` to `@section.blocks`, **not** an instance of `Comment`.

## Remedy

Registering collections for every descendant of a `reflection.klass` resolved this: https://github.com/stimulusreflex/cable_ready/compare/updatable-sti?expand=1#diff-90be2c3ebba84a166909a4f0528f7698ecea6f124f3d96c19a703e02320c38f9R93

**However** in the `development` and `test` environments, `Block.descendants` would return an empty array because of lazy loading. A ~~hack~~ compromise is to add an option `descendants:` to our `has_many` call:

https://github.com/stimulusreflex/cable_ready/compare/updatable-sti?expand=1#diff-90be2c3ebba84a166909a4f0528f7698ecea6f124f3d96c19a703e02320c38f9R93

https://github.com/stimulusreflex/cable_ready/compare/updatable-sti?expand=1#diff-90be2c3ebba84a166909a4f0528f7698ecea6f124f3d96c19a703e02320c38f9R44

**This actually doesn't do anything else than putting the class constant into ObjectSpace**. I.e. it doesn't matter where you call `Comment`, it just has to be before the collections are registered. Still, it feels like the most concise and documentable solution 💅 

EDIT: now explicitly `constantizing` descendants from an array of symbols or strings:

```rb
class Section
   has_many :blocks, enable_updates: true, descendants: [:Comment]
end
```


leastbad (commented):
So, I still appreciate this solution, as off-beat as it is and I'm basically good to go with it. I would ask that you append a tiny example of what a `has_many descendants: [Foo]` would look like, just for documentation posterity.

However, I do have a thought: what if we actually didn't throw out the descendants array and actually used it as a list on line 93? This would allow the developer to specify a list of classes that might exclude some.

The other thing that just popped into my head: is there such a thing as dynamically created STI classes? Not in typical operation, but I'm trying to think if there's a pattern which a popular gem might use that arbitrarily creates model classes.

Given autoloading in production, maybe that's not an issue, either. Just trying to black-hat this fix with the weirdest fuzzing ideas I can muster.

julianrubisch:
> what if we actually didn't throw out the descendants array and actually used it as a list on line 93? 

I actually thought the same. Just for consistency, would we want to go with a string/symbol specification instead, like class_name?

julianrubisch:
> trying to think if there's a pattern which a popular gem might use that arbitrarily creates model classes.

Hmm nothing that immediately springs to mind... every STI model has to write its type into the type column so... if we're talking about ActiveRecord, that would be quite backwards?

leastbad:
> > what if we actually didn't throw out the descendants array and actually used it as a list on line 93?
> 
> I actually thought the same. Just for consistency, would we want to go with a string/symbol specification instead, like class_name?

I think that your proposed syntax is already perfect. I was just hinting that it could be fed into the list as a splat, or such.

leastbad:
I feel like `descendants: [Comment]` is nicer than `descendants: [:Comment]`, no? And I thought the point was to get them to evoke the true name of their class so it can take its rightful place in the Object Space.

If `descendants: [Comment]` isn't possible, might I suggest `descendants: [:comment]` instead?

Fussy, perhaps. But consistency is underrated.

PR #208, opened 2022-05-31
Explicit check for leaf parent element presence for updates_for

rickbenavidez:
In certain scenarios, the call to recursive updates associated with updates_for may fail if the leaf's parent Element is null.  

This results in an error similar to the following:
`Uncaught TypeError: e2.parentElement is null`

To prevent the error we'll perform an explicit check on the presence of the parent element before attempting to call closest().


leastbad:
Would you mind running `yarn run prettier-standard:format` and commiting again, please?

rickbenavidez:
Thanks for the review @leastbad! Command run, commit amended.

PR #210, opened 2022-07-04
fixed 0 value problem with null checks

leastbad:
I applied @marcoroth's `value != null ? value : ''` to the CR operations which previously would have read `value || ''`. The issue was that if the developer sends `0` as a value, in JS `0` is falsey and so the OR falls over to `''`.

Some applications are more straightforward than others. Anything where a text/html value was being written, this is an obvious fix. However, in an attempt to be thorough, I also updated a few that I would appreciate 👀 on:

- `addCssClass`: can you have a class with the name `0`? I sort of doubt it.
- `pushState` and `replaceState`: should I be applying this to the `state` param? (I have it falling back to `{}`)
- `setCookie`: is this appropriate?
- `consoleTable`: is this appropriate?

### DIFF
--- a/javascript/operations.js
+++ b/javascript/operations.js
@@ -7,7 +7,10 @@ import {
   processElements,
   before,
   after,
-  operate
+  operate,
+  safeString,
+  safeArray,
+  safeObject
 } from './utils'
 
 export default {
@@ -18,7 +21,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { html, focusSelector } = operation
-        element.insertAdjacentHTML('beforeend', html != null ? html : '')
+        element.insertAdjacentHTML('beforeend', safeString(html))
         assignFocus(focusSelector)
       })
       after(element, operation)
@@ -45,7 +48,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { html, focusSelector } = operation
-        element.innerHTML = html != null ? html : ''
+        element.innerHTML = safeString(html)
         assignFocus(focusSelector)
       })
       after(element, operation)
@@ -57,10 +60,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { html, position, focusSelector } = operation
-        element.insertAdjacentHTML(
-          position || 'beforeend',
-          html != null ? html : ''
-        )
+        element.insertAdjacentHTML(position || 'beforeend', safeString(html))
         assignFocus(focusSelector)
       })
       after(element, operation)
@@ -72,10 +72,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { text, position, focusSelector } = operation
-        element.insertAdjacentText(
-          position || 'beforeend',
-          text != null ? text : ''
-        )
+        element.insertAdjacentText(position || 'beforeend', safeString(text))
         assignFocus(focusSelector)
       })
       after(element, operation)
@@ -115,7 +112,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { html, focusSelector } = operation
-        element.outerHTML = html != null ? html : ''
+        element.outerHTML = safeString(html)
         assignFocus(focusSelector)
       })
       after(parent.children[ordinal], operation)
@@ -127,7 +124,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { html, focusSelector } = operation
-        element.insertAdjacentHTML('afterbegin', html != null ? html : '')
+        element.insertAdjacentHTML('afterbegin', safeString(html))
         assignFocus(focusSelector)
       })
       after(element, operation)
@@ -153,7 +150,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { html, focusSelector } = operation
-        element.outerHTML = html != null ? html : ''
+        element.outerHTML = safeString(html)
         assignFocus(focusSelector)
       })
       after(parent.children[ordinal], operation)
@@ -165,7 +162,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { text, focusSelector } = operation
-        element.textContent = text != null ? text : ''
+        element.textContent = safeString(text)
         assignFocus(focusSelector)
       })
       after(element, operation)
@@ -179,7 +176,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { name } = operation
-        element.classList.add(...getClassNames(name != null ? name : ''))
+        element.classList.add(...getClassNames(safeString(name)))
       })
       after(element, operation)
     })
@@ -212,7 +209,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { name, value } = operation
-        element.setAttribute(name, value != null ? value : '')
+        element.setAttribute(name, safeString(value))
       })
       after(element, operation)
     })
@@ -223,7 +220,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { name, value } = operation
-        element.dataset[name] = value != null ? value : ''
+        element.dataset[name] = safeString(value)
       })
       after(element, operation)
     })
@@ -234,7 +231,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { name, value } = operation
-        if (name in element) element[name] = value != null ? value : ''
+        if (name in element) element[name] = safeString(value)
       })
       after(element, operation)
     })
@@ -245,7 +242,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { name, value } = operation
-        element.style[name] = value != null ? value : ''
+        element.style[name] = safeString(value)
       })
       after(element, operation)
     })
@@ -257,7 +254,7 @@ export default {
       operate(operation, () => {
         const { styles } = operation
         for (let [name, value] of Object.entries(styles))
-          element.style[name] = value != null ? value : ''
+          element.style[name] = safeString(value)
       })
       after(element, operation)
     })
@@ -268,7 +265,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { value } = operation
-        element.value = value != null ? value : ''
+        element.value = safeString(value)
       })
       after(element, operation)
     })
@@ -364,11 +361,7 @@ export default {
     before(window, operation)
     operate(operation, () => {
       const { state, title, url } = operation
-      history.pushState(
-        state != null ? state : {},
-        title != null ? title : '',
-        url
-      )
+      history.pushState(safeObject(state), safeString(title), url)
     })
     after(window, operation)
   },
@@ -413,11 +406,7 @@ export default {
     before(window, operation)
     operate(operation, () => {
       const { state, title, url } = operation
-      history.replaceState(
-        state != null ? state : {},
-        title != null ? title : '',
-        url
-      )
+      history.replaceState(safeObject(state), safeString(title), url)
     })
     after(window, operation)
   },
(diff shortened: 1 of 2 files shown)


javascript/operations.js
```diff
@@ -173,7 +179,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { name } = operation
-        element.classList.add(...getClassNames(name || ''))
+        element.classList.add(...getClassNames(name != null ? name : ''))
```
hopsoft, line 182: It appears that `getClassNames` is not returning what's expected. 

Search · getClassNames 2022-07-11 08-01-01

We may want to update it to something like this.

```js
const getClassNames = names => Array.from(names).flat()
```

Then update all usages to pass an Array argument on invocation.

```js
element.classList.add(...getClassNames([name != null ? name : '']))
```

javascript/operations.js
```diff
@@ -217,7 +223,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { name, value } = operation
-        element.dataset[name] = value || ''
+        element.dataset[name] = value != null ? value : ''
```
hopsoft, line 226: `dataset` assignment implicitly calls `toString` but I suppose forcing a string before invocation doesn't hurt.

javascript/operations.js
```diff
@@ -440,7 +454,7 @@ export default {
     operate(operation, () => {
       const { key, value, type } = operation
       const storage = type === 'session' ? sessionStorage : localStorage
-      storage.setItem(key, value || '')
+      storage.setItem(key, value != null ? value : '')
```
hopsoft, line 457: I'm pretty sure that storage assignment implicitly calls `toString` but forcing a string before invocation is 👍🏻

javascript/operations.js
```diff
@@ -452,8 +466,8 @@ export default {
     operate(operation, () => {
       const { message, level } = operation
       level && ['warn', 'info', 'error'].includes(level)
-        ? console[level](message || '')
-        : console.log(message || '')
+        ? console[level](message != null ? message : '')
+        : console.log(message != null ? message : '')
```
hopsoft, line 470: Technically `console.log` can accept any data type. I wonder if we should omit type coercion for this one.

javascript/operations.js
```diff
@@ -473,7 +487,8 @@ export default {
       const { title, options } = operation
       Notification.requestPermission().then(result => {
         operation.permission = result
-        if (result === 'granted') new Notification(title || '', options)
+        if (result === 'granted')
+          new Notification(title != null ? title : '', options)
```
hopsoft, line 491: We should probably coerce `options` to an object for consistency.

hopsoft (changes requested):
Given that we're applying type coercion in quite a few places now, I think adding some utility functions defined in `utils.js` would make sense and help keep things consistent.

- `toString`
- `toArray`
- `toObject`

leastbad:
Amazing feedback! Will tackle after a sleep cycle.

javascript/operations.js
RolandStuder: It does, but `null` get set to `"null"` (string) rather than blank. Which we may or may not want.

javascript/operations.js
RolandStuder: I agree:
`cable_ready.console_log(message: nil})` will be communicating better that it received `nil` rather than a empty string

javascript/operations.js
```diff
@@ -462,7 +476,7 @@ export default {
     before(document, operation)
     operate(operation, () => {
       const { data, columns } = operation
-      console.table(data, columns || [])
+      console.table(data, columns != null ? columns : [])
```
RolandStuder, line 479: Irrelevant, but TIL decimal values are converted to a string, when using `to_json`

javascript/operations.js
RolandStuder: If something is passed in, that is not an object, it gives a JS error currently, as somewhat cryptic one:

```
[Error] CableReady detected an error in notification: undefined is not an object (evaluating 'Notification.requestPermission().then'). If you need to support older browsers make sure you've included the corresponding polyfills. https://docs.stimulusreflex.com/setup#polyfills-for-ie11.
	(anonymous function) (vendors-node_modules_hotwired_turbo-rails_app_javascript_turbo_index_js-node_modules_rails_ac-d86665-82aec27bfd70959b3050.js:8773:7788)
	forEach
	q (vendors-node_modules_hotwired_turbo-rails_app_javascript_turbo_index_js-node_modules_rails_ac-d86665-82aec27bfd70959b3050.js:8773:7275)
	O (vendors-node_modules_ahoy_js_dist_ahoy_esm_js-node_modules_animejs_lib_anime_es_js-node_modul-b9f7a7-9fc50a36129a03869516.js:49384:9039)
	map
	message
	```
Not sure how you would coerce something into a string if someone does:
`cable_ready.notification(title:"hello", options: :foo)`

Maybe it's better to let it blow up at that point, or to report a specific error that it can't convert it.

File: javascript/operations.js [hunk 7]
```diff
@@ -18,7 +18,7 @@ export default {
       before(element, operation)
       operate(operation, () => {
         const { html, focusSelector } = operation
-        element.insertAdjacentHTML('beforeend', html || '')
+        element.insertAdjacentHTML('beforeend', html != null ? html : '')
```
RolandStuder, line 21: Curiosity: why did you not go for the `??`
https://caniuse.com/mdn-javascript_operators_nullish_coalescing

```js
        element.insertAdjacentHTML('beforeend', html && '')
```

RolandStuder (commented):
I think what you did is the appropriate way to go.

Generally we want operations to change strings, but yes you not a few exceptions: but as far as I could tell, there isn't really a situation where casting to a blank string goes against expectations (except for the console.log one).

As @hopsoft said, in some interesting situations like :set_storage_item and  :set_dataset_property there would be an implicit conversion to a string, but I don't think we want the implicit conversion as `nil` will be cast to `"null"` (string).

So in my mind the following question popped up: "is there a situation where I explicitly want "null" to be the value. I think that use case is a possibility, though no one has complained so far :-), but I think it's way less surprising to default to a blank string than to a string "null", if somebody really wants it they cut broadcast "null" (string).

So I think this is all good.

javascript/operations.js
leastbad: I don't see value in the *string* `'null'`. If someone wants it, they should pass `"null"` as a string.

javascript/operations.js
leastbad: I went with `safeString`, `safeArray` and `safeObject` because `toString` is already a thing in JS.

With these new methods, we have an opportunity, in @RolandStuder's vision, to potentially do a bit better than just failing with an exception.

Right now, we have
```js
function safeObject (obj) {
  return obj != null ? obj : {}
}
```
Option 1: do nothing, we're done, let people own their mistakes
Option 2: we could test `obj` to verify that it is an `Object`, and if not, we could log an error that tells people they are passing the wrong thing
Option 3: we could test `obj` to verify that it is an `Object`, and if not, we could silently replace the bad value with `{}`
Option 4: we could test `obj` to verify that it is an `Object`, and if not, we could log an error that tells people they are passing the wrong thing AND coerce the value to `{}` so that events fire and etc

What do you both like best? Obviously I'd do the same thing for all three functions.

javascript/operations.js
leastbad: That strikes me as super WTF but outside of our scope or ability to fix. Right?

javascript/operations.js
leastbad: Ha! I wish I had a dollar....

TL;DR: esbuild seems to choke on it depending on vague things. Can't trust it. Same with safe navigation. Will revisit topic in ~2025.

javascript/utils.js
```diff
@@ -130,6 +130,18 @@ function handleErrors (response) {
   return response
 }
 
+function safeString (str) {
+  return str != null ? str : ''
```
hopsoft, line 134: Should we explicitly cast in these helper functions?

```js
return str != null ? String(str) : ''
```

javascript/utils.js
```diff
@@ -130,6 +130,18 @@ function handleErrors (response) {
...
+}
+
+function safeArray (arr) {
+  return arr != null ? arr : []
```
hopsoft, line 138:
```js
return arr != null ? Array.from(arr) : []
```

javascript/utils.js
```diff
@@ -130,6 +130,18 @@ function handleErrors (response) {
...
+}
+
+function safeObject (obj) {
+  return obj != null ? obj : {}
```
hopsoft, line 142:
```js
return obj != null ? Object(obj) : {}
```

hopsoft (changes requested):

javascript/utils.js
leastbad: Super smart.

hopsoft (approved):

leastbad:
@hopsoft that just leaves https://github.com/stimulusreflex/cable_ready/pull/210#discussion_r922886469

javascript/utils.js
```diff
@@ -131,14 +131,26 @@ function handleErrors (response) {
 }
 
 function safeString (str) {
+  if (typeof str !== 'string') {
+    console.warn(`Operation requires a String, but got ${str} (${typeof str})`)
+    return ''
+  }
   return str != null ? String(str) : ''
```
RolandStuder, line 138: No, I don't like it. This means if I transmit my operation that should set an attribute to the integer of 1. It will allow this operation server side, but then on client side will cast it to `''`, which is not what I would expect.

This may actually break existing apps, that transmits Integers to dataset value, where they relied on JS default behavior to cast it to a string.

Considering where my thing came from
```ruby
set_dataset_property(selector: "[data-add-to-cart-item-gid-value=\"#{@purchasable.to_gid_param}\"]",
      name: "addToCartQuantityValue",
      value: quantity,
      select_all: true)
```

where quantity was an integer, it worked fine, except for when it was `0`, now it would fail for all numbers, unless I cast it `to_s` first server side. So I would expect to be automatically cast either server side, when creating the operation or client side.

leastbad: Alright, thanks to quality logic review from @RolandStuder I believe that the correct course of action is to introduce a 4th function, `safeScalar`. It warns if the value passed is not `['string', 'number', 'boolean'].includes(typeof val)` and doesn't mess with the value passed, unless the value passed is null or undefined (which gets corrected to `''`).

I've done a pass through the operations to update them to use `safeString` and `safeScalar` where appropriate. For example, when operating on CSS class names or property names, `safeString` is applied. (CSS classes shouldn't start with numbers, even though many browsers seem to allow it for compatibility.) For the values themselves, `safeScalar`.

The one scenario that is more complex is actually `morph`, since it works on a `template` element. What I did here was replace `String(html).trim()` with `String(safeScalar(html)).trim()` because some folks might conceivably want to send a `0` to morph.

I didn't attempt to analyze or mutate any `url` values eg. `go`, `pushState`, `redirectTo` because I don't want to get into actual validation. Open to opinions here.

Anyhow, what do you think of the solution as presented? Is `string`, `number`, `boolean` the complete list?

hopsoft: I like `safeScalar` and think that `string`, `number`, and `boolean` cover our use cases of what will emit from the server. Also, the logic outlined above for URLs etc... seems reasonable to me, but I also think expecting a String or even Scalar for URLs is "good enough" for our needs.

leastbad: I just pushed a commit that protects `url` options with `safeString`.

leastbad: BTW I don't think that a literal number can be interpreted as a valid URL.

Interestingly, even with `666` in my `/etc/hosts`, `ping 666` resolves to `0.0.2.154`. Of course `ping 667` resolves to `0.0.2.155`.

TL;DR: scalar is not necessary.

PR #222, opened 2022-10-21
remove strict dependencies on Active Record, Action Cable and Active Storage

### LINKED ISSUE #207: ActiveRecord dependency breaks rspec on non AR Rails implementations
I have an app running with StimulusReflex ~4.x~  3.4 fine, this app have no AR whatsoever, as we use other DB which AR do not support. The migration to 5.0pre is making RSpec tests crash because AR is being required for the new identifiable feature.

I think this only happens with RSpec on test environment, as when setting eager load to true in my dev env the app works fine. 

Thanks for the awesome work!

leastbad:

Enhancement

## Description

In an effort to be helpful, I've manually reimplemented the changes in 828c7d073ca149ca293d989dbf1ac8a2304783f9 22399b3ee26ad631916cc062894a5cbead525289 and 24c2f0bb520543bade0aaaafaccbc09bc0908203. That is, @marcoroth wrote 100% of the code in this commit. I just wanted to help get the changes in, quickly. I'm confident that there's a fancy git cherry pick command that would do this, but I also needed to make sure it merged cleanly and since there's not that much to this PR, I just wanted to get it done. If there's a better way, close this PR without delay!

Concerns before merging:

1. Are there new/additional things which need to be covered? I admit that I'm not 100% sure I understand *why* these changes were all necessary, which means I can't guarantee that any new requirements are accounted for... if any.
2. Do we need to do similar optimizations in SR?

Fixes #207 

## Why should this be added

Tidying up loose ends!

## Checklist


leastbad:
`standardrb` failed in a strange way that I do not understand: https://github.com/stimulusreflex/cable_ready/actions/runs/3294765236/jobs/5432607555

marcoroth:
> `standardrb` failed in a strange way that I do not understand

I guess it might be because of this line: https://github.com/stimulusreflex/cable_ready/pull/222/files#diff-2f4da7b98a9c86d38f368f6119614e27dbb9c1e4dc3d70240cb466cc211435acR24

and/or that the standard action is still running on Ruby 2.6

leastbad:
I don't know how to change how actions are run.

https://github.com/stimulusreflex/cable_ready/blob/master/.github/workflows/standardrb.yml doesn't seem to specify a Ruby version. Is there a setting somewhere for this? Nothing about runtime environment on https://github.com/stimulusreflex/cable_ready/settings/actions

marcoroth:
> https://github.com/stimulusreflex/cable_ready/blob/master/.github/workflows/standardrb.yml doesn't seem to specify a Ruby version. Is there a setting somewhere for this? Nothing about runtime environment on https://github.com/stimulusreflex/cable_ready/settings/actions

I was being set by the third party GitHub Action we were using. I resolved this via #223.

But I think the tests failures we are seing now are legit because of some missing ActiveRecord requires?

lib/cable_ready/identifiable.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "action_view"
```
marcoroth, line 3: I guess we need to require `active_record` here and probably also define `action_view` and `active_record` as dependencies of the gem.

lib/cable_ready/identifiable.rb
```diff
@@ -21,10 +23,26 @@ def dom_id(record, prefix = nil)
     end
 
     def identifiable?(obj)
-      obj.respond_to?(:to_dom_selector) ||
+      (
+        obj.respond_to?(:to_dom_selector) ||
         obj.respond_to?(:to_dom_id) ||
-        obj.is_a?(ActiveRecord::Relation) ||
-        obj.is_a?(ActiveRecord::Base)
+        ar_relation?(obj) ||
```
javierg, line 29: the tests fails here because how inheritance in ruby will prevent methods on the CableReady to be accesible when method is included in a nested includes.

As in https://github.com/stimulusreflex/cable_ready/pull/222/files#diff-bb8e7da8f683e6a053f852e827078c740fd88706ae7ca90dfbb515ef417b09ecR7

this module is nested inside, when calling ar_reliation? it is not there so you get the error.... I think 🤔

leastbad:
How do we move this one forward, @marcoroth? Are you able to give this a few minutes? It would be great to cross this issue off the list.

leastbad:
Alright, after a substantial amount of pain, I appear to have sorted out the issues preventing the tests and CI steps from passing.

Thanks to @javierg and @KonnorRogers for the assists!

I'd like to renew the push to either get this fixed and merged, or close the chapter. To reiterate, the goal of this PR is to make it possible to address #207 by making it possible to use CR in a non-Active Record project.

If this goal is possible, then we can document which steps need to be taken (in terms of requires etc) and which features are off-limits. I have never used anything but Active Record, and I barely understand zeitwerk, so I am woefully ill-prepared to take this PR over the finish line.

If this goal is not possible, we will have to break the news and this PR should be closed with regrets. @marcoroth, can you take over responsibility for this PR and the decisions motivating it? I've gone about as far as I can with it.

leastbad:
@javierg I don't want to jinx it, but I *think* that I have CableReady working properly without ActiveRecord in scope. It shows up in `Gemfile.lock` a few times, as a dependency of frameworks like ActiveStorage. However, I believe that I have <b>*</b>everything working without AR being `require`d.

WindowsTerminal_U0Yl6Q0V0S

1. `rails new foo -j esbuild -O`
2. add `"cable_ready", github: "stimulusreflex/cable_ready", branch: "new_installer"` to `Gemfile`
3. `rails cable_ready:install` and accept all defaults (hit enter ~5 times)
4. change `Gemfile` so that the `cable_ready` branch is `dependency_cleanup`
5. `bin/dev` in one Terminal tab
6. open `localhost:3000/example` in browser
7. `rails c` in a second Terminal tab
8. execute these lines<br><code>include CableReady::Broadcaster<br>cable_ready[:example_page].text_content("#stream_from_output", text: "Hello from the console!").broadcast</code>
9. You should see the message you sent in your browser, and you can execute `defined?(ActiveRecord)` and see it's not required.

@marcoroth I am still feeling pretty fuzzy about all of this; it's not 100% clear to me what all of the shuffling that wasn't AR-specific was intended to do, or if you even finished this task before getting distracted. Some code review is appreciated, though FWIW the tests are passing.

Don't sweat the conflicts; I'll handle all of the touch-ups when it comes time to merge.

Anyhow, @javierg I am so sorry that this took so long, and I hope that it still helps. Can you please give it a few tire kicks and let us know if anything seems off?

<b>*</b> The big caveat for this PR is that while the `Identifiable` module works just fine without Active Record, the `Updatable` functionality is disabled. If you try to include `CableReady::Updatable` in a non-AR class, it will 100% blow up. That's not something we can address at this time, sorry!

javierg:
@leastbad thank you for this work, I tested your branch on a project were AR is not present and it works as expected in the basic escenarios we have.

PR #224, opened 2022-11-13
Rename Helper to ViewHelper so Rails will include it correctly

erlingur:
# Bug fix

## Description

Recently @marcoroth namespaced the helper file under the CableReady namespace (#220). This broke using `stream_from` view helper since Rails seems to prefer a helper be called xxxHelper. I've made a suggestion to rename Helper to ViewHelper. I'm very open to other names though.

## Why should this be added

This makes the `stream_from`, `updates_for` and other view helpers work again.

### DIFF
--- a/app/helpers/cable_ready/view_helper.rb
+++ b/app/helpers/cable_ready/view_helper.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module CableReady
-  module Helper
+  module ViewHelper
     include CableReady::Compoundable
     include CableReady::StreamIdentifier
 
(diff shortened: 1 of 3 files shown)


leastbad:
LGTM!

Some day, I'll understand how the whole force-pushing thing works in practice. 😸

erlingur:
> Some day, I'll understand how the whole force-pushing thing works in practice. 😸

I wanted the commit that said "Fix tests" to *actually* fix the tests 🙈

marcoroth (approved):
Thanks @erlingur! I guess Rails expects the file to be called `*_helper.rb` in order to be auto-loaded.

PR #225, opened 2022-11-19
fix `updates_for` makes redundant calls

### LINKED ISSUE #209: `updates_for` Makes Redundant Calls
I think the order in which an `updates_for_element` checks whether it `shouldUpdate` may be a little off.

Let's say I have the following html:
```html.erb
<%= updates_for @item, only: :foo do %>
  ...
<% end %>
```

And then I update a field not specified in the `only` option:
```rb
@item.update(bar: true)
```

If we look at the [update(data)](https://github.com/stimulusreflex/cable_ready/blob/ebc9c3d60fe00963ab8bed44515521ac94b1861a/javascript/elements/updates_for_element.js#L48) method of `updates_for_element.js`, the logic will play out in the following order:

1. We create a new `Block` out of the element (L49).
2. We fetch the html for the element (L65). This part is the problem, I think. Ideally, we shouldn't be requesting the html until we know the element should actually be updated, but we haven't run that check yet.
3. We process the block (L85). It's only now that we realize the element shouldn't be updated, and so we don't proceed with the actual updating. But this means the fetch in step `2.` above was redundant: we asked the server to render the full html for the page but didn't do anything with it.

I think the better approach here would be to check whether the update should be performed _before_ we go and pester the server, so something along the lines of:

```js
let blocks = Array.from(
  document.querySelectorAll(this.query),
  element => new Block(element)
)

// filter out any blocks that don't require updates
blocks = blocks.filter(block => block.shouldUpdate(data))

// return here if there are no blocks that require updates
if (blocks.length === 0) return

// carry on with the fetch logic only for blocks that require updates
...
```

Not sure if I'm missing something here, but I played around with the above locally and it seemed to work well.

@julianrubisch @leastbad penny for your thoughts :)

leastbad:

Bug fix

## Description

Modify `updates_for` element to verify that a page is required before fetching it.

This PR essentially implements the suggestion made in #209, with minor tweaks.

I've done my best to test this new functionality, but I am not a heavy user of `updates_for` and would appreciate if @julianrubisch could sanity check this. Specifically, my concern is that there could be an unexpected interaction with lazily loaded Turbo Frame content?

I was concerned that filtering out blocks might cause the index-based element selection to point to the wrong element, but so far, I haven't been able to trip it up. I really don't want to break this functionality.

@andrewerlanger I also removed the previous line 117, which was where we performed the `shouldUpdate` check initially. It seems as though this should no longer be necessary. Is there a good reason to run the check again?

I've tested on a simple User model with `enable_updates`, as well as `name` and `email` attributes:
```erb
<%= updates_for current_user, only: :email do %>
  <p><%= current_user.name %></p>
  <p><%= current_user.email %></p>
<% end %>

<%= updates_for current_user, only: :name do %>
  <p><%= current_user.name %></p>
  <p><%= current_user.email %></p>
<% end %>
```
Seems to work fine when I update either attribute.

Fixes #209 

## Why should this be added

@andrewerlanger pointed out that we fetch pages from the server before checking if they are used, which is a waste of computing resources and bandwidth.

## Checklist

### DIFF
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -55,10 +55,10 @@ export default class UpdatesForElement extends SubscribingElement {
     const blocks = Array.from(
       document.querySelectorAll(this.query),
       element => new Block(element)
-    )
+    ).filter(block => block.shouldUpdate(data))
 
     // first updates-for element in the DOM *at any given moment* updates all of the others
-    if (blocks[0].element !== this) return
+    if (blocks.length === 0 || blocks[0].element !== this) return
 
     // hold a reference to the active element so that it can be restored after the morph
     ActiveElement.set(document.activeElement)
@@ -113,9 +113,6 @@ class Block {
   }
 
   async process (data, html, index) {
-    // with the index incremented, we can now safely bail - before a fetch - if there's no work to be done
-    if (!this.shouldUpdate(data)) return
-
     const blockIndex = index[this.url]
     const template = document.createElement('template')
     this.element.setAttribute('updating', 'updating')


julianrubisch (commented):
This looks like a sane approach 👍🏻

I'll see if I can check this against an app of mine.

Would be nice if @andrewerlanger did the same, provided you still use it.

leastbad:
@julianrubisch @andrewerlanger any feedback so far? I'd love to get this merged if everything looks okay.

andrewerlanger:
@julianrubisch @leastbad not sure how I missed these messages but only just seeing them now. Sorry for the radio silence and many thanks for pushing ahead with the fix :)

PR #232, opened 2022-12-18
Fix bug where using an array when calling (add/remove)_css_class didn…

### LINKED ISSUE #231: Using an array with addCssClass does not update the classList correctly
While using cable_ready's `addCssClass` with an array of class names, the classList of the element(s) are updated incorrectly.
They are updated, but the class names are separated by a comma (,) instead of a space.

Eg:
Using
```ruby
add_css_class(selector: '#foo', name: ["btn-warning", "pulsating"])
```
will update the '#foo' element's classList with `btn-warning,pulsating` instead of `btn-warning pulsating`

Additional info:
This error message is shown in the console
warning

binarygit:
This is a bugfix for #231 , while the original issue only mentions `add_css_class` while working on this, I realized that the same bug affected `remove_css_class` and so fixed that too.

## Description

Please include a summary of the change and which issue is fixed.

Fixes #231 

## Why should this be added
It fixes a bug.

## Checklist

Please note that the best way to suggest changes or updates to the [documentation](https://docs.stimulusreflex.com) is to [join Discord](https://discord.gg/stimulus-reflex) and leave a note in the #docs channel. Any documentation updates posted as PRs cannot be accepted at this time. :heart:

### DIFF
--- a/javascript/operations.js
+++ b/javascript/operations.js
@@ -11,7 +11,8 @@ import {
   safeScalar,
   safeString,
   safeArray,
-  safeObject
+  safeObject,
+  safeStringOrArray
 } from './utils'
 
 export default {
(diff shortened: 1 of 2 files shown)


leastbad (approved):
Great work on this.

I added a quick check to remove the `class` attribute if there are no classes defined.

PR #244, opened 2023-01-31
Add setting switch to disable assets compilation

n-rodriguez:
Like ActionCable: https://github.com/rails/rails/blob/main/actioncable/lib/action_cable/engine.rb#L27 

Useful when using Webpacker to manage static assets

- [ ] My code follows the style guidelines of this project
- [ ] Checks (StandardRB & Prettier-Standard) are passing
- [ ] This is not a documentation update

### DIFF
--- a/lib/cable_ready/config.rb
+++ b/lib/cable_ready/config.rb
@@ -11,7 +11,7 @@ module CableReady
     include Observable
     include Singleton
 
-    attr_accessor :on_failed_sanity_checks, :on_new_version_available
+    attr_accessor :on_failed_sanity_checks, :on_new_version_available, :precompile_assets
     attr_writer :verifier_key
 
     def initialize
@@ -19,6 +19,7 @@ module CableReady
       @operation_names = Set.new(default_operation_names)
       @on_failed_sanity_checks = :exit
       @on_new_version_available = :ignore
+      @precompile_assets = true
     end
 
     def observers
--- a/lib/cable_ready/engine.rb
+++ b/lib/cable_ready/engine.rb
@@ -2,6 +2,21 @@ require "rails/engine"
 
 module CableReady
   class Engine < Rails::Engine
+    # If you don't want to precompile CableReady's assets (eg. because you're using webpack),
+    # you can do this in an initializer:
+    #
+    # config.after_initialize do
+    #   config.assets.precompile -= CableReady::Engine::PRECOMPILE_ASSETS
+    # end
+    PRECOMPILE_ASSETS = %w[
+      cable_ready.js
+      cable_ready.min.js
+      cable_ready.min.js.map
+      cable_ready.umd.js
+      cable_ready.umd.min.js
+      cable_ready.umd.min.js.map
+    ]
+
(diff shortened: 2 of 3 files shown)


marcoroth:
Thanks @n-rodriguez! I thought we had that already, but seems like we don't!

I'm wondering if we should do it like `turbo-rails` does it?

https://github.com/hotwired/turbo-rails/blob/6db05d69d5f060bd142defe247550b5da5c47228/lib/turbo/engine.rb#L23-L35

n-rodriguez:
I'd prefer to not touch [CableReady::Config](https://github.com/stimulusreflex/cable_ready/blob/master/lib/cable_ready/config.rb) so let's do it à la Turbo :)

marcoroth:
> I'd prefer to not touch `CableReady::Config`

@n-rodriguez Do you have any specific reason for that?

n-rodriguez:
> @n-rodriguez Do you have any specific reason for that?

to keep things simple

n-rodriguez:
Actually I don't use the `main` branch of CableReady, I use `5.0.0.pre9` for both Ruby and JS so I'm not aware of the changes in the APIs.

Maybe this is better :

```ruby
# config/initializers/cable_ready.rb
CableReady.configure do |config|
  config.precompile_assets = false # new configuration accessor

  config.on_failed_sanity_checks = :ignore

  config.add_operation_name :notify
  config.add_operation_name :progress_bar
end
```
but it may requires some tests. wdyt?

n-rodriguez:
FWIW :

Turbo : https://github.com/hotwired/turbo-rails/blob/main/lib/turbo/engine.rb#L31
ActionText: https://github.com/rails/rails/blob/main/actiontext/lib/action_text/engine.rb#L33
ActionCable: https://github.com/rails/rails/blob/main/actioncable/lib/action_cable/engine.rb#L27
ActiveStorage: https://github.com/rails/rails/blob/main/activestorage/lib/active_storage/engine.rb#L189

marcoroth:
@n-rodriguez I like the version where we have it in `config/initializers/cable_ready.rb` 👍🏼 

but we can also keep the `PRECOMPILE_ASSETS` constant so you could also do it the other way if you like.

For ActionText, ActionCable and ActiveStorage it makes sense that they do it that way because the already have their regular a config on `application.config`

marcoroth:
One more thing: Would you mind adding it commented out to: https://github.com/stimulusreflex/cable_ready/blob/29b7e063465acb0468ecd3d4034b0c00696c7f07/lib/generators/cable_ready/templates/config/initializers/cable_ready.rb#L3 

Then we can merge this, imo 🚀

n-rodriguez:
> Then we can merge this, imo rocket

Do you think it's possible to cut a new pre-release for both Ruby and JS?

marcoroth:
Yes, pre10 is going out this week!

marcoroth (approved):
Thank you @n-rodriguez!

n-rodriguez:
> Yes, pre10 is going out this week!

Thanks!

PR #255, opened 2023-02-14
fix: dom_id from view helper

julianrubisch:
# Bug Fix

## Description

Fixes the broken `dom_id` helper that was cause by `ViewHelper` including `Broadcaster`, which in turn included `Identifiable`, which contains a patched version of dom_id

## Checklist

### DIFF
--- a/app/helpers/cable_ready/view_helper.rb
+++ b/app/helpers/cable_ready/view_helper.rb
@@ -4,7 +4,6 @@ module CableReady
   module ViewHelper
     include CableReady::Compoundable
     include CableReady::StreamIdentifier
-    include CableReady::Broadcaster
 
     def stream_from(*keys, html_options: {})
       tag.stream_from(**build_options(*keys, html_options))
@@ -25,6 +24,10 @@ module CableReady
       else
         capture(&block)
       end
+    end
+
+    def cable_car
+      CableReady::CableCar.instance
     end
 
     private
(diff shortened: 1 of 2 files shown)


app/helpers/cable_ready/view_helper.rb
```diff
@@ -27,6 +26,10 @@ def updates_for_if(condition, *keys, **options, &block)
       end
     end
 
+    def cable_car
+      CableReady::CableCar.instance
+    end
+
```
marcoroth, line 32: Does it even make sense that `cable_car` is available in the View scope? What would you use this for?

test/lib/cable_ready/view_helper_test.rb
```diff
@@ -3,8 +3,6 @@
 require "test_helper"
 
 class CableReady::ViewHelperTest < ActionView::TestCase
-  include CableReady::Broadcaster
-
```
marcoroth, line 7: We probably want to include the `CableReady::ViewHelper` instead here

julianrubisch: It is included. Verified with a failing test

app/helpers/cable_ready/view_helper.rb
julianrubisch: No clue but I don’t wanna break stuff

julianrubisch: Probably some folks use it in controllers etc?

julianrubisch: Well, I don’t know 

Anyway I don’t wanna remove stuff without understanding the ramifications

marcoroth (approved):

PR #259, opened 2023-02-25
Namespace Broadcastjob

julianrubisch:
# Enhancement

## Description

Puts the `BroadcastJob` in the `CableReady` namespace.

## Why should this be added

To line it up with all the other modules.

## Checklist


marcoroth (approved):
This is good! 

I was wondering about this the last time too when we were namespacing all the things, but then intentionally left this one out because I thought it might break already enqueued jobs. 

But now after thinking about this again, I think it makes sense because these jobs shouldn't be enqueued over a long period of time anyway.

PR #260, opened 2023-03-01
Rework CableReady's `ActionController::Renderers`

marcoroth:
# Type of PR

Deprecation + Enhancement

## Description

CableReady's `render operations:` ActionController renderer doesn't indicate that it's coming/being used for CableReady operations. This pull request renames the renderer to `render cable_ready:` and adds a deprecation warning for the `operations` renderer.

Additionally it make the DX around it a little bit better by checking if the passed argument responds to `.dispatch` before calling it. This helps when you already called `.dispatch` in the controller action or if you provide it a already converted JSON string.

This pull request also changes the `cable_ready` MIME type from `application/vnd.cable-ready.json` to `text/vnd.cable-ready.json` so that the response from a controller action can be viewed in a browser.

Finally, this pull request adds the `X-Cable-Ready-Version` header to the response. 

Screenshot 2023-03-04 at 14 11 07

## Why should this be added

This is an effort to keep the naming consistent and scope all CableReady functions with its own name. Following up on #259, #252 and #250.

## Checklist

### DIFF
--- a/lib/cable_ready/engine.rb
+++ b/lib/cable_ready/engine.rb
@@ -41,16 +41,15 @@ module CableReady
           response.content_type ||= Mime[:cable_ready]
           response.headers["X-Cable-Ready-Version"] = CableReady::VERSION
 
-          render json: operations.dispatch
+          render json: operations.respond_to?(:dispatch) ? operations.dispatch : operations
         end
 
         ActionController::Renderers.add :cable_ready do |operations, options|
           response.content_type ||= Mime[:cable_ready]
           response.headers["X-Cable-Ready-Version"] = CableReady::VERSION
 
-          render json: operations.dispatch
+          render json: operations.respond_to?(:dispatch) ? operations.dispatch : operations
         end
-
       end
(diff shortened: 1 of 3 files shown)


lib/cable_ready/engine.rb
```diff
@@ -29,14 +29,27 @@ class Engine < Rails::Engine
...
-          render json: operations.dispatch
+          response.headers["X-Cable-Ready-Version"] = CableReady::VERSION
+
+          render json: operations.respond_to?(:dispatch) ? operations.dispatch : operations
         end
 
-        Mime::Type.register "application/vnd.cable-ready.json", :cable_ready
+        ActionController::Renderers.add :cable_ready do |operations, options|
+          response.content_type ||= Mime[:cable_ready]
+          response.headers["X-Cable-Ready-Version"] = CableReady::VERSION
```
hopsoft, line 49: The addition of the version header will be helpful.

hopsoft (approved):

PR #261, opened 2023-03-01
Debounced updatable

julianrubisch:
# Enhancement

## Description

One of the big downsides of using `CableReady::Updatable` in the wild is that it can lead to a lot of noise on the ActionCable connection. This is because every change on a model that enables updates leads to a ping being sent over ActionCable due to the usage of ActiveRecord callbacks.

This PR introduces a `debounce:` option that takes an `ActiveSupport::Duration` argument that will only send out one ping for this model per debounce period.

It makes use of a simple internal per-process `MemoryDebounceAdapter`, which is just a thin synchronized layer around a `Hash`.

Preparations are made, however, to override this adapter using a `CableReady::Updatable.debounce_adapter` mattr_accessor, which can be overridden in an initializer. This should make it possible to devise other adapters (for example based on DRb or Layered Caching) in the future.

## Checklist

### DIFF
--- a/app/models/concerns/cable_ready/updatable.rb
+++ b/app/models/concerns/cable_ready/updatable.rb
@@ -206,21 +206,23 @@ module CableReady
         return if skip_updates_classes.any? { |klass| klass >= self }
         raise("ActionCable must be enabled to use Updatable") unless defined?(ActionCable)
 
-        @debounce_time ||= 0.seconds
-
-        if @debounce_time > 0.seconds
+        if debounce_time > 0.seconds
           key = compound([model_class, *options])
           old_wait_until = CableReady::Updatable.debounce_adapter[key]
           now = Time.now.to_f
 
           if old_wait_until.nil? || old_wait_until < now
-            new_wait_until = now + @debounce_time.to_f
+            new_wait_until = now + debounce_time.to_f
             CableReady::Updatable.debounce_adapter[key] = new_wait_until
             ActionCable.server.broadcast(model_class, options)
           end
         else
           ActionCable.server.broadcast(model_class, options)
         end
+      end
+
+      def debounce_time
+        @debounce_time ||= 0.seconds
       end
 
       def skip_updates_classes
--- a/app/models/concerns/cable_ready/updatable.rb
+++ b/app/models/concerns/cable_ready/updatable.rb
@@ -8,11 +8,19 @@ module CableReady
 
     class MemoryDebounceAdapter
       include Singleton
+      include MonitorMixin
 
-      delegate_missing_to :@keys
+      delegate_missing_to :@dict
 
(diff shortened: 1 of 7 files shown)


app/models/concerns/cable_ready/updatable.rb
```diff
@@ -6,6 +6,26 @@ module CableReady
...
+      def initialize
+        super
+        @dict = {}
+      end
+
+      def []=(key, value)
+        synchronize do
+          @dict[key] = value
+        end
+      end
```
hopsoft, line 24: Should we synchronize reads also? I dont' think we're quite ready to pull in a new dependency but we should at least keep an eye out for options like [`Concurrent::Hash`](https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/Hash.html) as we chart our way into this territory.

app/models/concerns/cable_ready/updatable.rb
```diff
@@ -30,11 +50,14 @@ def self.enable_cable_ready_updates(*options)
           options = options.extract_options!
           options = {
             on: [:create, :update, :destroy],
-            if: -> { true }
+            if: -> { true },
+            debounce: 0.seconds
```
hopsoft, line 54: I wonder if it would make sense to have the default provide some degree of protection. Perhaps 100 milliseconds?

hopsoft (changes requested):
This is looking great.

app/models/concerns/cable_ready/updatable.rb
julianrubisch: yep... I was looking at https://github.com/hopsoft/coin/blob/master/lib/coin/vault.rb, and since you don't synchronize reads there it confirmed my gut feeling that it shouldn't be necessary. But I'll happily add it if you think it's necessary.

app/models/concerns/cable_ready/updatable.rb
julianrubisch: well I switch on the debounce time in `broadcast_updates` to gracefully allow the default behavior. But sure, why not!

julianrubisch:
Also note that this doesn't yet factor in the case of `enable_updates` on associations (that would be next on my list)

hopsoft:
Will association support land in a separate PR?

julianrubisch:
> Will association support land in a separate PR?

I thought it would be wiser to split it, yes

app/models/concerns/cable_ready/updatable.rb
julianrubisch: actually I just noticed that this breaks all the tests, so we'd have to specify `debounce: 0` everywhere. I'd recommend merging this PR and then introducing this in a separate one to keep things readable

PR #262, opened 2023-03-01
Fix inverse of resolution

### LINKED ISSUE #258: `inverse_of` returns `nil` when enriching `has_many` association with updates
## `inverse_of` returns `nil`
I'm running into a strange issue where `inverse_of` returns `nil` although I have declared it in the model.

For example, I have two models, `Listing` which has many `Listings::PublishAction`s.
```ruby
# Listing
has_many :publish_actions, class_name: "Listings::PublishAction", dependent: :destroy, foreign_key: :listing_id, enable_updates: true, inverse_of: :listing

# Listing::PublishAction
belongs_to :listing, class_name: "Listing"
```

Inside of [enrich_association_with_updates](https://github.com/stimulusreflex/cable_ready/blob/4d0cd016af504b886b4a72e14a5eab25486dc244/app/models/concerns/cable_ready/updatable.rb#L112), this line gets the `inverse_of` reflection:

```
inverse_of = reflection.inverse_of&.name&.to_s
```

This is returning `nil` for me, and I get an error from the [collection registry](https://github.com/stimulusreflex/cable_ready/blob/f9228a13706814f7871fbbed927eb0a489f0524e/app/models/concerns/cable_ready/updatable/collections_registry.rb#L25).

## `inverse_name` has the proper value

Although this is `nil`, `inverse_name` has the proper key, and the options have the `inverse_of` declaration inside of it as well:

```
[5] pry(Listing)> self
=> Listing(id: integer, team_id: integer, name: string, created_at: datetime, updated_at: datetime)
[6] pry(Listing)> name
=> :publish_actions
[7] pry(Listing)> reflect_on_association(name)
=> #<ActiveRecord::Reflection::HasManyReflection:0x00007f4b1e4d97f0
 @active_record=
  Listing(id: integer, team_id: integer, name: string, created_at: datetime, updated_at: datetime),
 @class_name="Listings::PublishAction",
 @inverse_name=:listing,
 @inverse_of=nil,
 @klass=
  Listings::PublishAction(id: integer, listing_id: integer, started_at: datetime, completed_at: datetime, target_count: integer, performed_count: integer, scheduled_for: datetime, sidekiq_jid: string, created_by_id: integer, approved_by_id: integer, created_at: datetime, updated_at: datetime),
 @name=:publish_actions,
 @options=
  {:class_name=>"Listings::PublishAction",
   :dependent=>:destroy,
   :foreign_key=>:listing_id,
   :inverse_of=>:listing,
   :extend=>[ExtendHasMany::ClassMethods, ObfuscatesId::ClassMethods]},
 @plural_name="publish_actions",
 @scope=nil>
[8] pry(Listing)> reflect_on_association(name).inverse_of
=> nil
```

This doesn't happen however when firing up the Rails console and grabbing the reflection that way:

```
> rails c
Loading development environment (Rails 7.0.4.2)
irb(main):001:0> Listing.reflect_on_association(:publish_actions).inverse_of
=> 
#<ActiveRecord::Reflection::BelongsToReflection:0x00007f50ddbdb098                                
 @active_record=Listings::PublishAction (call 'Listings::PublishAction.connection' to establish a connection),
 @klass=nil,                                                                                      
 @name=:listing,                                                                                  
 @options={:class_name=>"Listing"},                                                               
 @plural_name="listings",                                                                         
 @scope=nil>   
```

Also, everything works by simply removing `enable_updates: true`, which led me to believe it's something going on here. For what it's worth, it's happening in a Bullet Train application, but it doesn't seem like it's something there. I wish I could give more details, but that's about all I can glean from the code right now.

julianrubisch:
# Bug Fix

## Description

This is more or less a reimplementation of a patch that @andrewculver made to fix `inverse_of` resolution. I just took it a bit further and lazy loaded/stored reflection properties in the finder, because it's useless to do the work twice, and declutters the main module a bit

https://github.com/stimulusreflex/cable_ready/compare/master...andrewculver:cable_ready:master?diff=split

(maybe) Fixes #258 
awaiting confirmation from the BulletTrain team cc @andrewculver

## Checklist

### DIFF
--- a/app/models/concerns/cable_ready/updatable/collections_registry.rb
+++ b/app/models/concerns/cable_ready/updatable/collections_registry.rb
@@ -37,14 +37,14 @@ module CableReady
       def find_resource_for_update(collection, model)
         collection.reflection ||= collection.klass.reflect_on_association(collection.name)
 
-        collection.foreign_key ||= collection.reflection.foreign_key
+        collection.foreign_key ||= collection.reflection&.foreign_key
 
         # lazy load and store through and inverse associations
         if collection.reflection&.through_reflection?
-          collection.inverse_association ||= collection.reflection.through_reflection.inverse_of&.name&.to_s
-          collection.through_association = collection.reflection.through_reflection.name.to_s.singularize
+          collection.inverse_association ||= collection.reflection&.through_reflection.inverse_of&.name&.to_s
+          collection.through_association = collection.reflection&.through_reflection.name.to_s.singularize
         else
-          collection.inverse_association ||= collection.reflection.inverse_of&.name&.to_s
+          collection.inverse_association ||= collection.reflection&.inverse_of&.name&.to_s
         end
 
         raise ArgumentError, "Could not find inverse_of for #{collection.name}" unless collection.inverse_association
--- a/app/models/concerns/cable_ready/updatable/collections_registry.rb
+++ b/app/models/concerns/cable_ready/updatable/collections_registry.rb
@@ -2,6 +2,17 @@
 
 module CableReady
   module Updatable
+    Collection = Struct.new(
+      :klass,
+      :name,
+      :reflection,
+      :options,
+      :foreign_key,
+      :inverse_association,
+      :through_association,
+      keyword_init: true
+    )
+
     class CollectionsRegistry
       def initialize
         @registered_collections = []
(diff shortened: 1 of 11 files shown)


gazayas:
This solves #258 for me :bow:

app/models/concerns/cable_ready/updatable/collections_registry.rb
```diff
@@ -12,23 +23,35 @@ def register(collection)
...
         end
       end
 
       private
 
       def find_resource_for_update(collection, model)
-        raise ArgumentError, "Could not find inverse_of for #{collection[:name]}" unless collection[:inverse_association]
+        collection.reflection ||= collection.klass.reflect_on_association(collection.name)
+
+        collection.foreign_key ||= collection.reflection.foreign_key
```
hopsoft, line 40: We use safe navigation below on `collection.reflection`. Should we do the same here?

hopsoft (approved):

PR #263, opened 2023-03-01
Set default debounce time in CableReady::Config

julianrubisch:
# Enhancement

## Description

Allows to set the default `updatable_debounce_time` in the CableReady config. This allows every application to set it to a desired value

## Checklist

### DIFF
--- a/app/models/concerns/cable_ready/updatable.rb
+++ b/app/models/concerns/cable_ready/updatable.rb
@@ -57,7 +57,7 @@ module CableReady
           options = {
             on: [:create, :update, :destroy],
             if: -> { true },
-            debounce: 0.seconds
+            debounce: CableReady.config.updatable_debounce_time
           }.merge(options)
 
           enabled_operations = Array(options[:on])
@@ -228,7 +228,7 @@ module CableReady
       end
 
       def debounce_time
-        @debounce_time ||= 0.seconds
+        @debounce_time ||= CableReady.config.updatable_debounce_time
       end
 
       def skip_updates_classes
--- a/app/models/concerns/cable_ready/updatable.rb
+++ b/app/models/concerns/cable_ready/updatable.rb
@@ -212,13 +212,13 @@ module CableReady
         return if skip_updates_classes.any? { |klass| klass >= self }
         raise("ActionCable must be enabled to use Updatable") unless defined?(ActionCable)
 
-        if debounce_time > 0.seconds
+        if debounce_time > 0
           key = compound([model_class, *options])
           old_wait_until = CableReady::Updatable.debounce_adapter[key]
           now = Time.now.to_f
 
           if old_wait_until.nil? || old_wait_until < now
-            new_wait_until = now + debounce_time.to_f
+            new_wait_until = now + debounce_time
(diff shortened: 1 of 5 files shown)


marcoroth (commented):
@julianrubisch looks good to me! Could you add a test for it? I recently added https://github.com/stimulusreflex/cable_ready/pull/254 which should be super straightforward.

julianrubisch:
> @julianrubisch looks good to me! Could you add a test for it? I recently added #254 which should be super straightforward.

I wanted to, but it introduces a race condition which sometimes breaks updatable tests 😕

marcoroth:
Could we resolve that with a similar helper method we added in StimulusReflex?

https://github.com/marcoroth/stimulus_reflex/blob/dcb7048f60611185f34bda4744e4e81e1000d40e/test/configuration_test.rb#L5-L11

julianrubisch:
I’ll give it a try!

julianrubisch:
Splendid! 🙌 

@hopsoft this also allows us to leave the default debounce time at 0.1 seconds

app/models/concerns/cable_ready/updatable.rb
```diff
@@ -228,7 +228,7 @@ def broadcast_updates(model_class, options)
       end
 
       def debounce_time
-        @debounce_time ||= 0.seconds
+        @debounce_time ||= CableReady.config.updatable_debounce_time
```
hopsoft, line 231: Should we coerce the value type with a `to_f` here?

hopsoft (changes requested):

app/models/concerns/cable_ready/updatable.rb
julianrubisch: hmmm actually that's done in line 221 where it makes more sense?

hopsoft: But 215 is expecting a numeric `if debounce_time > 0.seconds`. Just trying to guard against people doing silly stuff.

hopsoft: I guess setting it to `nil` could be considered a valid use case depending on the individual? 🤷🏻‍♂️

julianrubisch: gotcha 👍

lib/generators/cable_ready/templates/config/initializers/cable_ready.rb
```diff
@@ -19,4 +19,9 @@
   # Change the default Active Job queue used for broadcast_later and broadcast_later_to
   #
   # config.broadcast_job_queue = :default
+
+  # Specify a default debounce time for CableReady::Updatable callbacks
+  # Doing so is a best practice to avoid heavy ActionCable traffic
+  #
+  # config.updatable_debounce_time = 0.1.seconds
```
marcoroth, line 26: @julianrubisch I think we should also add this to:

https://github.com/stimulusreflex/stimulus_reflex/blob/main/lib/generators/stimulus_reflex/templates/config/initializers/cable_ready.rb

julianrubisch: 👍🏻

PR #264, opened 2023-03-03
Updatable debounce associations

julianrubisch:
# Enhancement

## Description

This is an extension of the debounce mechanism to associations. Note that I switched to passing the debounce time around via method arguments rather than instance variables, because it was cleaner to do so from both the model class as well the association extension side.

There may be some opportunities to declutter this a bit, though at the moment they're not obvious to me.

## Checklist

### DIFF
--- a/app/models/concerns/cable_ready/updatable.rb
+++ b/app/models/concerns/cable_ready/updatable.rb
@@ -62,11 +62,9 @@ module CableReady
 
           enabled_operations = Array(options[:on])
 
-          @debounce_time = options[:debounce]
-
-          after_commit(ModelUpdatableCallbacks.new(:create, enabled_operations), {on: :create, if: options[:if]})
-          after_commit(ModelUpdatableCallbacks.new(:update, enabled_operations), {on: :update, if: options[:if]})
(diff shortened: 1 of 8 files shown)


app/models/concerns/cable_ready/updatable.rb
```diff
@@ -151,12 +156,13 @@ def enrich_association_with_updates(name, option, descendants = nil)
...
             options: options,
-            reflection: reflection
+            reflection: reflection,
+            debounce_time: debounce
           ))
         end
       end
 
-      def enrich_attachments_with_updates(name, option)
+      def enrich_attachments_with_updates(name, option, debounce: 0.seconds)
```
hopsoft, line 165: Should the default for `debounce` here match the other method signatures? i.e.

```ruby
debounce: CableReady.config.updatable_debounce_time
```

hopsoft (changes requested):

app/models/concerns/cable_ready/updatable.rb
julianrubisch: Oops, that slipped through…

PR #266, opened 2023-03-03
Make memory debounce adapter use MemoryCacheStore

julianrubisch:
# Enhancement

## Description

Makes the memory debounce adapter use `ActiveSupport::MemoryCacheStore` inherently, with a default expiration time of 5 minutes.

Note that I also moved it to a new adapters directory in anticipation of other adapters.

## Checklist

### DIFF
--- a/lib/cable_ready.rb
+++ b/lib/cable_ready.rb
@@ -17,6 +17,8 @@ require "cable_ready/version"
 require "cable_ready_helper"
 
 module CableReady
+  autoload :MemoryCacheDebounceAdapter, "cable_ready/updatable/memory_cache_debounce_adapter"
+
   class << self
     def config
       CableReady::Config.instance
--- a/lib/cable_ready/updatable/memory_cache_debounce_adapter.rb
+++ b/lib/cable_ready/updatable/memory_cache_debounce_adapter.rb
@@ -1,0 +1,22 @@
+module CableReady
+  module Updatable
+    class MemoryCacheDebounceAdapter
+      include Singleton
+
+      delegate_missing_to :@store
+
+      def initialize
+        super
+        @store = ActiveSupport::Cache::MemoryStore.new(expires_in: 5.minutes)
+      end
+
+      def []=(key, value)
+        @store.write(key, value)
+      end
+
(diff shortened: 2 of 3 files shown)


lib/cable_ready.rb
```diff
@@ -17,6 +17,8 @@
 require "cable_ready_helper"
 
 module CableReady
+  autoload :MemoryCacheDebounceAdapter, "cable_ready/updatable/memory_cache_debounce_adapter"
```
marcoroth, line 20: Is there any reason why this is using `autoload` over a regular `require`?

julianrubisch: the app at hand might not use `Updatable`, which is itself autoloaded by living in `app/`

julianrubisch: also, in the future there will probably be more than one adapter which you can specify via config. then even more it wouldn't make sense to preemptively load all of them.

lib/cable_ready/updatable/memory_cache_debounce_adapter.rb
```diff
@@ -0,0 +1,22 @@
+module CableReady
+  module Updatable
+    class MemoryCacheDebounceAdapter
+      include Singleton
+
+      delegate_missing_to :@store
+
+      def initialize
+        super
+        @store = ActiveSupport::Cache::MemoryStore.new(expires_in: 5.minutes)
```
hopsoft, line 10: > This cache has a bounded size specified by the :size options to the initializer (default is 32Mb).

I'd recommend that we limit max size even further to something like 8Mb or 16Mb. At the very least, I'd like to make max size explicit so it's more obvious. Thoughts?

hopsoft (changes requested):
This is fantastic. Had one minor request.

lib/cable_ready/updatable/memory_cache_debounce_adapter.rb
julianrubisch: Yeah totally. I also thought to make it configurable, but at present this is probably an overdo.

I think 8Mb will do.

PR #274, opened 2023-03-30
Eager load MemoryCacheDebounceAdapter

julianrubisch:
# Bugfix

## Description

autoloading MemoryCacheDebounceAdapter intermittently caused `uninitialized constant CableReady::Updatable::MemoryCacheDebounceAdapter` as reported by @obie and others. back to eager loading it.

## Checklist


Matt-Yorkley:
I'm still getting weirdness with this, I've tried applying these changes but the problem just shifts to a slightly different one which gives me intermittent errors saying `enable_cable_ready_updates` is undefined...

Slapping a `require "cable_ready/updatable/memory_cache_debounce_adapter"` at the top of the `CableReady::Updatable` concern and just calling it by the class name seems to do the trick:

```ruby
mattr_accessor :debounce_adapter, default: MemoryCacheDebounceAdapter.instance
```

PR #276, opened 2023-05-08
Add trigger and target element logs for devtools

julianrubisch:
# Enhancement

## Description

Furnishes cable-ready-updates-for-element with instance variables for element logs, so they can displayed in the devtools like this:

CleanShot 2023-05-08 at 12 22 44@2x

### DIFF
--- a/javascript/elements/updates_for_element.js
+++ b/javascript/elements/updates_for_element.js
@@ -61,11 +61,16 @@ export default class UpdatesForElement extends SubscribingElement {
       element => new Block(element)
     ).filter(block => block.shouldUpdate(data))
 
-    this.triggerElementLog.push(Log.request(data, blocks))
+    this.triggerElementLog.push(
+      `${new Date().toLocaleString()}: ${Log.request(data, blocks)}`
+    )
 
     if (blocks.length === 0) {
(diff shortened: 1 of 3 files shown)


javascript/elements/updates_for_element.js
```diff
@@ -32,6 +32,9 @@ export default class UpdatesForElement extends SubscribingElement {
     super()
     const shadowRoot = this.attachShadow({ mode: 'open' })
     shadowRoot.innerHTML = template
+
+    this.triggerElementLog = []
+    this.targetElementLog = []
```
hopsoft, line 37: Do we need to clear these logs or perhaps have a log stack that we move them to when `cable-ready:after-update` is dispatched?

hopsoft (changes requested):

javascript/elements/updates_for_element.js
julianrubisch: 🤔

You are probably correct, though in my tests it didn't really matter because they are cleared with every browser refresh.

I think the easiest would be to limit them to a small number (10?) and do a kind of FIFO thing so they get cleared automatically?

hopsoft (approved):

PR #277, opened 2023-05-28
Fix lingering class-loading issues with CableReady::Updatable concern

Matt-Yorkley:
# Bugfix

## Description

There's still some weirdness with the way the `MemoryCacheDebounceAdapter` is being loaded in `CableReady::Updatable` concern. Notes on the previous fix here: https://github.com/stimulusreflex/cable_ready/pull/274#issuecomment-1565767491

It seems to be resolved now, I'm not seeing the intermittent errors after this tweak :+1:

## Checklist

### DIFF
--- a/app/models/concerns/cable_ready/updatable.rb
+++ b/app/models/concerns/cable_ready/updatable.rb
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 
 require "active_support/concern"
+require "cable_ready/updatable/memory_cache_debounce_adapter"
 
 module CableReady
   module Updatable
     extend ::ActiveSupport::Concern
 
-    mattr_accessor :debounce_adapter, default: ::CableReady::Updatable::MemoryCacheDebounceAdapter.instance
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
I'm wondering in which cases the class loading issues are showing up though. I haven't seen them arise in my personal apps, but this looks good to me!

Any concerns @julianrubisch?

julianrubisch (approved):
No concerns, but it seems we'll want to move the whole module over to `lib`.

Really weird, but also quite the outside my horizon 🤷‍♂️

Matt-Yorkley:
Ah! I think I was switching between too many branches when testing and one of them was still on `5.0.0-rc2` so the `MemoryCacheDebounceAdapter` was still under `/lib`... I'll take another look.

Matt-Yorkley:
Yeah I was on the wrong branch. Sorry for the false alarm!

marcoroth:
Thanks for doubling-checking, @Matt-Yorkley!

PR #278, opened 2023-05-30
Set xpath flag if/when overwriting selector with previous_selector

Matt-Yorkley:
# Bug fix

## Description

When chaining operations that use a `selector` argument, subsequent operations can optionally omit the `selector` argument and the `selector` from the previous operation gets applied applied automatically (at line 46 of `CableReady::OperationBuilder` [here](https://github.com/stimulusreflex/cable_ready/blob/main/lib/cable_ready/operation_builder.rb#L46).

There's a weird edge case here that can cause the subsequent operation to fail if the first operation was using `xpath`, because the `selector` argument gets set but the `xpath` argument does not get set and defaults to `false`, eg:
```ruby
cable_ready
  .add_css_class(name: "example", selector: "/html/body/div[1]", xpath: true) # succeeds
  .inner_html(html: "fail") # fails
```
The first operation is applied correctly but the `inner_html` operation fails to find it's element, as it has `selector: "/html/body/div[1]", xpath: false`, resulting in: 

```
CableReady innerHtml operation failed due to missing DOM element for selector: '/html/body/div[1]`
```
The DOM element isn't missing, it's just not being looked for with `xpath`.

## Why should this be added

Fixes an obscure edge case when chaining operations using xpath selectors and omitting the selector in subsequent operations.

## Checklist

### DIFF
--- a/lib/cable_ready/operation_builder.rb
+++ b/lib/cable_ready/operation_builder.rb
@@ -3,7 +3,7 @@
 module CableReady
   class OperationBuilder
     include Identifiable
-    attr_reader :identifier, :previous_selector
+    attr_reader :identifier, :previous_selector, :previous_xpath
 
     def self.finalizer_for(identifier)
       proc {
@@ -45,10 +45,11 @@ module CableReady
         options["selector"] = selector if selector && options.exclude?("selector")
         if previous_selector && options.exclude?("selector")
           options["selector"] = previous_selector
-          options["xpath"] = true if options["selector"].to_s.starts_with? "/"
+          options["xpath"] = previous_xpath if previous_xpath
         end
         if options.include?("selector")
           @previous_selector = options["selector"]
+          @previous_xpath = options["xpath"]
           options["selector"] = identifiable?(previous_selector) ? dom_id(previous_selector) : previous_selector
         end
         options["operation"] = name.to_s.camelize(:lower)
--- a/lib/cable_ready/operation_builder.rb
+++ b/lib/cable_ready/operation_builder.rb
@@ -43,7 +43,10 @@ module CableReady
           options.each { |key, value| options[key] = value.send("to_#{key}".to_sym) if value.respond_to?("to_#{key}".to_sym) }
         end
         options["selector"] = selector if selector && options.exclude?("selector")
-        options["selector"] = previous_selector if previous_selector && options.exclude?("selector")
+        if previous_selector && options.exclude?("selector")
+          options["selector"] = previous_selector
+          options["xpath"] = true if options["selector"].to_s.starts_with? "/"
+        end
         if options.include?("selector")
           @previous_selector = options["selector"]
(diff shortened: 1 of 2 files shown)


lib/cable_ready/operation_builder.rb
```diff
@@ -43,7 +43,10 @@ def add_operation_method(name)
           options.each { |key, value| options[key] = value.send("to_#{key}".to_sym) if value.respond_to?("to_#{key}".to_sym) }
         end
         options["selector"] = selector if selector && options.exclude?("selector")
-        options["selector"] = previous_selector if previous_selector && options.exclude?("selector")
+        if previous_selector && options.exclude?("selector")
+          options["selector"] = previous_selector
+          options["xpath"] = true if options["selector"].to_s.starts_with? "/"
```
marcoroth, line 48: I'm wondering if it would make sense to also store the `xpath` option in a `previous_xpath` variable instead of basing it off of the `options["selector"]` value.

And even more meta, now that I'm thinking about it: couldn't we just leave out this `xpath` option all-together for all operations and check on the client-side if any given selector starts with a `/`? Regular CSS selectors shouldn't start with a slash.

marcoroth (commented):
Thank you for taking this one on @Matt-Yorkley!

lib/cable_ready/operation_builder.rb
Matt-Yorkley: Maybe both? It feels a bit weird if the server side is building operations that aren't valid and passing them to the client anyway?

marcoroth: Why wouldn't they be valid? 

Maybe we can keep it simple for now and just store the previous xpath value and leave it at. 

But it might be something to think about down the road.

Matt-Yorkley: I dropped the last commit :ok_hand:

marcoroth: Awesome, thank you!

lib/cable_ready/operation_builder.rb
```diff
@@ -43,9 +43,13 @@ def add_operation_method(name)
           options.each { |key, value| options[key] = value.send("to_#{key}".to_sym) if value.respond_to?("to_#{key}".to_sym) }
         end
         options["selector"] = selector if selector && options.exclude?("selector")
-        options["selector"] = previous_selector if previous_selector && options.exclude?("selector")
+        if previous_selector && options.exclude?("selector")
+          options["selector"] = previous_selector
+          options["xpath"] = previous_xpath if previous_xpath
```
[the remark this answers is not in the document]
Matt-Yorkley: I think I added that in because the output looked a bit weird otherwise. Namely; in the case where `previous_xpath` is `nil`, it would assign the `xpath` key to the payload, but with an undefined value. It felt like not assigning it at all and not increasing the payload size was preferable in that case? Like if `xpath` is `nil`, then the payload shouldn't have an `xpath` key at all instead of having an empty one? What do you think?

PR #280, opened 2023-06-06
Debounce adapter config option

julianrubisch:
# Enhancement

## Description

This allows to set a different debounce adapter than the built-in one, in preparation to building one with https://github.com/hopsoft/composite_cache_store/

Initially I wanted to move the `MemoryCacheDebounceAdapter` to `lib/cable_ready/updatable/memory_cache_debounce_adapter.rb`, but that again resulted in weird Zeitwerk issues when requiring it that led nowhere. That's why I opted into defining the default adapter in a config hook.

Eventually we'll want to move the whole updatable cod from `app/` to `lib/`.

## Checklist

### DIFF
--- a/app/models/concerns/cable_ready/updatable.rb
+++ b/app/models/concerns/cable_ready/updatable.rb
@@ -1,12 +1,11 @@
 # frozen_string_literal: true
 
 require "active_support/concern"
+require "cable_ready/config"
 
 module CableReady
   module Updatable
     extend ::ActiveSupport::Concern
-
-    mattr_accessor :debounce_adapter, default: ::CableReady::Updatable::MemoryCacheDebounceAdapter.instance
 
     included do |base|
       if defined?(ActiveRecord) && base < ActiveRecord::Base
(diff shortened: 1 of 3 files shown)


app/models/concerns/cable_ready/updatable.rb
```diff
@@ -1,13 +1,12 @@
 # frozen_string_literal: true
 
 require "active_support/concern"
+require "cable_ready/config"
```
hopsoft, line 4: I'd consider exposing an accessor on the CableReady module similar to `signed_stream_verifier` rather than referencing the debounce adapter through `CableReady.config`. Thoughts?

hopsoft (changes requested):

julianrubisch:
Hmmm only that it wouldn't fit into the current initializer template, which would be kinda weird?

app/models/concerns/cable_ready/updatable.rb
julianrubisch: Hmmm only that it wouldn't fit into the current initializer template, which would be kinda weird?

IIRC the signed_stream_verifier is considered private API whereas this ... is not?

Another thing is documentation, which I need to improve

hopsoft: Just verified that referencing `config` in multiple places is the strategy I'm using in Turbo Boost libs too. Carry on. 🖖🏻

hopsoft (approved):

PR #285, opened 2023-09-14
Wrap updates-for-element in an IntersectionObserver

julianrubisch:
# Feature

## Description

This PR addresses an optimization regarding the number of requests that `Updatable` possibly makes.

The rationale is this: it doesn't really make sense to keep elements that aren't visible up to date. this can be deferred, which defaults to lazy loading.

We have to distinguish 3 cases here:

- is visible - behavior as usual
- is not visible - ignore requests. Basically shouldUpdate == false
- transition from invisible to visible - pull and morph once, then listen as normal

## Implementation

I leaned on Turbo's solution here a bit, and introduced a separate `AppearanceObserver`. The thing I added to the regular `IntersectionObserver` is an observer for `document.visibilityState`. This means:

- not only hidden elements in the DOM tree of the opened browser tab
- but also generally elements that are in browser tabs that are currently closed/hidden

will _not_ trigger updates. When they come into view, e.g. by scrolling into the viewport, or switching browser tabs, they will perform a one-time update and then continue listening as normal.

In terms of view helpers, I only added an `observe_appearance` flag to keep this optional. The element will then observe both behaviors though.

### DIFF
--- a/app/helpers/cable_ready/view_helper.rb
+++ b/app/helpers/cable_ready/view_helper.rb
@@ -27,13 +27,13 @@ module CableReady
       tag.cable_ready_stream_from(**build_options(*keys, html_options))
     end
 
-    def cable_ready_updates_for(*keys, url: nil, debounce: nil, only: nil, ignore_inner_updates: false, observe_intersection: false, html_options: {}, &block)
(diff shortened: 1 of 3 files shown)


hopsoft:
This type of optimization is 🤌  Very nice work here.

hopsoft (approved):
Bravo! 👏🏻

PR #300, opened 2024-08-04
Add `cable_ready.integration_test_request_encoding` initializer

tcannonfodder:
# Type of PR

Enhancement

## Description

* This has been adapted from https://github.com/hotwired/turbo-rails/blob/9b17a3be3705786d72c3ae77fde5a9b3006555d7/lib/turbo/engine.rb#L119
* In order to streamline development, Rails integration tests should be able to support `as: :cable_ready` to test Cable Ready responses
	* This is done by setting up a new request encoding that ActionDispatch's integration tests can use. In this case, the encoder is barebones and is largely focused on streamlining how the request `Accept` headers are setup

## Why should this be added

This makes it easier for folks to write tests for their CableReady actions, making the gem fit into the larger ecosystem

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] Checks (StandardRB & Prettier-Standard) are passing


lib/cable_ready/engine.rb
```diff
@@ -49,6 +49,20 @@ class Engine < Rails::Engine
       end
     end
 
+    initializer "cable_ready.integration_test_request_encoding" do
+      ActiveSupport.on_load(:action_dispatch_integration_test) do
+        # Support `as: :cable_ready`. Public `register_encoder` API is a little too strict.
+        class ActionDispatch::RequestEncoder
+          class CableReadyStreamEncoder < IdentityEncoder
+            header = [ Mime[:cable_ready], Mime[:html] ].join(",")
```
marcoroth, line 57: Hey @tcannonfodder, thanks for this pull request. Should this be `:json` instead of `:html`, since the MIME-Type is `text/vnd.cable-ready.json`:

```suggestion
            header = [ Mime[:cable_ready], Mime[:json ].join(",")
```

marcoroth: Yes, you are totatally @kaspth! Thanks!

kaspth:
@marcoroth heyo, anything I can do to help on this?

marcoroth:
Hey @kaspth thanks for checking in! I couldn't apply the fix myself, but if @tcannonfodder could to update the suggestion that would be great! Then I'm more than happy to merge this! Thanks! 🙌🏼

tcannonfodder:
Will do this weekend! Sorry, things have been a mess here

marcoroth:
Thanks @tcannonfodder, looks like standard is failing on your changes, if you could run the fix command and push up the changes that would be super appreciated! 🙌🏼

tcannonfodder:
I _think_ I've got standardrb appeased (I am having issues with `nio4r` on my M1 Mac, so a bit blind here)

kaspth:
Nice!

kaspth:
I just looked at the latest RequestEncoder setup in Rails and I wonder if we should do something more akin to https://github.com/rails/rails/blob/20f5828cacf164f0f53d669b37ba1e4b0f0f8907/actionpack/lib/action_dispatch/testing/request_encoder.rb#L58 here.

Ref: https://github.com/rails/rails/commit/0f4ab82082c87f22b6733d46b4251009cbbb1dd4

That way users can match on the operations JSON:

```ruby
patch account_url(@account), params: { name: "Yo" }, as: :cable_ready

assert_pattern { response.parsed_body => {name:} } # Not quite sure what the test would look like, I'm not the most experienced with CableReady.
```

We could also do a custom response parser object, that would load the JSON, construct into objects that respond to `deconstruct` & `deconstruct_keys` and then parse the internal HTML via `Rails::Dom::Testing.html_document.parse`.

We could start with:

```ruby
@encoders[:cable_ready] = CableReadyStreamEncoder.new \
  response_parser: ->(body) { JSON.parse(body, object_class: ActiveSupport::HashWithIndifferentAccess)
```

Also because of the way the request encoders are setup, ^ would only be called when the response acutally was a `cable_ready` response. An HTML fallback response would correctly use the HTML appropriate response parser.

kaspth:
cc @seanpdoyle in case you have thoughts on extending the request encoder.

PR #306, opened 2025-01-24
Update link for mrujs

louim:
Update the documentation with the (I assume) correct link for `mrujs`.


marcoroth (approved):
Thanks @louim, this looks right! Seems like the old domain is not in @KonnorRogers hands anaymore 🙈

